PR #23, opened 2013-10-18
Poltergeist 1

rewritten:
There are two specs that fail, but the same code is good in console. The specs roll back, but nothing seems to trigger the rollback.

@naoisegolden @amartincaro @enricostano Could you help me please? :microscope:


markets:
@coopdevs/timeoverflow-contributors Creo que también se puede cerrar.

PR #113, opened 2014-12-23
Tasks admin tags

danielmacho72:
Refactored code to avoid extra code in _grouped_index.html.haml including test to verify
Added main logic and bootstrap style (based on checkboxes) to allow merging and deleting tags if user is admin

3 new views: 

1 refactored previous views to only show tags for normal users to view 
2 to allow admin users for tag management in their organization (1 view to delete another to rename)

Added control to avoid merges without new names and with no checkbox selected.
Added warning popup before merging or deletion

Still Pending for future PR:
  Adding Tests

Ready to be merged in master and deployed in preprod


app/views/tags/_grouped_admin_checkboxes.html.haml
```diff
@@ -0,0 +1,20 @@
+%div
+  - i=0
```
rewritten: What's `i` for?

app/views/tags/_grouped_admin_checkboxes.html.haml
```diff
@@ -0,0 +1,20 @@
...
+  - alpha_tags.each_slice(6) do |gr|
```
rewritten: No need to slice, just use a column layout! Using the autoprefixer it's as easy as
```css
.six-columns {
  column-count: 6;
}
```
and put the boxes one under the other.

rewritten:
This is how you do it without angularjs.
```coffeescript
$("#tag_form").validate
  rules:
    "tags[]": required: true
```
the plugin already disables any submit action on a non-valid form so there is absolutely nothing more to do, no controller, no specific attributes on each checkbox, and no actual browser memory used for a scope as the validation plug-in only executes on input changes.

Just add the `rails-assets-jquery-validate` gem to the gemfile and `#= require jquery-validate` in the main app.js.coffee file.

danielmacho72:
I've tried to use your proposal for several hours to avoid using angularJS but the issue using jquery-validation-rails (https://github.com/danryan/jquery-validation-rails) is that:

1) Only allows one locale per application, you have to load their messages
2) If you try to override the messages it won't allow you to use #{t "xxxx"} to specify your locale translated message per each of required fields but only a fixed text again in one given language

So for me using this https://github.com/danryan/jquery-validation-rails is not an option until they do not allow multiple locales per application.

I am keeping AngularJS ng-disabled strategy unless you have some other better idea.

app/views/tags/_grouped_admin_checkboxes.html.haml
danielmacho72: I've removed slice, no need to use column-count but I have been forced to change style as follows to avoid weird layout when having several lines for a given cell an next one only one, then instead of creating a new row it was just putting it below. 

Reviewing in Internet I found that you have to clear to force a new row and maintain the responsiveness

Try with this style, you'll notice 6 columns and reducing size of screen is respecting responsiveness for smaller size properly. Otherwise you get a weird layout.

under tags.css.scss:
.col-md-2:nth-child(7)  {
	clear: both;
}

rewritten:
Creo que esta pull request es demasiado compleja. Hay demasiadas vistas, demasiado código, los tests de controller solamente comprueban que se renderice cierto template pero sin comprobar que el template realmente funciona, hay una serie de sugerencias que se han ignorado completamente.

Por mi no se puede hacer merge.

Probemos de otra manera: vamos a hacer una simple pull request, con solamente un formulario para borrar tags (el rename ya lo haremos después). Sin AngularJS. Sin un layout demasiado complejo. Lo más importante es la mantenibilidad.

PR #143, opened 2015-03-05
added missing tests for Member

carlesjove:
Added a bunch of tests for `Member`. Also some edit on the Gemfile just to keep it cleaner.

NOTE: Pending test. After some attempts, I thought that i was missing something. Please, if you can provide me some feedback of how an offer is really built, I'll fix it and push back.


spec/models/member_spec.rb
```diff
@@ -1,8 +1,37 @@
...
+  it { should delegate_method(:balance).to(:account).with_prefix }
+  it { should delegate_method(:gender).to(:user).with_prefix }
+  it { should delegate_method(:date_of_birth).to(:user).with_prefix }
+
+  it { should validate_presence_of(:organization_id) }
+  it { should validate_presence_of(:member_uid) }
+
+  pending "#offers" do
+    let(:member) { Fabricate(:member) }
+    let(:offer) { Fabricate(:offer, user: member.user) }
```
rewritten: This should work:
```
let(:offer) { Fabricate(:offer, user: member.user, organization: member.organization) }
```
A post is diamond-linked to the member, as it is from a user, and it is "inside" an organization.

rewritten:
Re: the Gemfile, we initially used Faker to do some sort of demo seeds, thus it was out of the :test scope. But now we don't anymore. Thanks for the cleanup.

spec/models/member_spec.rb
carlesjove: I had tried that, already. The problem is that `member.organitzation` returns the name of the organization, tot the object. Anyway, I'll dig a little more and see what I can do.

rewritten: That is a bug.

spec/spec_helper.rb
```diff
@@ -9,6 +9,7 @@
 # in spec/support/ and its subdirectories.
 Dir[Rails.root.join("spec/support/**/*.rb")].each { |f| require f }
 
+ActiveRecord::Migration.maintain_test_schema!
```
carlesjove: I'd like to have this change in a separate commit, but it seems that we currently need this in order to run tests. Needs a fix at some point, though.

carlesjove:
Done!

spec/spec_helper.rb
rewritten: If you add it in a separate PR, I'll accept and merge. Then we can merge master back here and it's separated.

carlesjove: OK. I'll move it to its own commit, wait until it's merged, and push back after a rebase.

markets: @carlesjove I'm not 100% sure, but I think we can remove `ActiveRecord::Migration.check_pending!` if we use `ActiveRecord::Migration.maintain_test_schema!`, could you please check it? Thanks.

carlesjove: @markets Sure! Gimme a minute

rewritten: True.

rewritten: I was too quick to merge.

rewritten: Why not regenerating the whole RSpec boilerplate from scratch?

carlesjove: Well, it's no big deal. We can leave it as it is, a rm it in another commit, for reference.

According to the docs, since Rails 4.1 it's no longer necessary: https://github.com/rails/rails/blob/ce392d948403a4b5cb7348669cb597a3c93b8d3c/guides/source/upgrading_ruby_on_rails.md#changes-to-test-helper

carlesjove: I'm going to rebase and push back, if it's ok by you

rewritten: :+1:

carlesjove:
Done. Unless you think there's something else to be done here, I think it's ready

rewritten:
Thanks Carles

PR #146, opened 2015-03-09
Add validation of Org web format

rewritten:
Replaces #122 allowing the field to be `nil`.


markets:
@rewritten :+1: por mi

PR #148, opened 2015-03-19
redirect back to current page after sign in and add info text on puclic ...

PatriRor:
...posts


markets:
@PatriRor por mi :+1:

PR #172, opened 2015-09-26
Switch avatars from robohash to avatars.io [WIP]

rewritten:
With cucumber feature specifications.


rewritten:
@markets @carlesjove :heart: :heart: :heart:

d2bit:
Some suggestions:
Why don't you add `Cucumber` in a separate PR?
On the other hand, untracking editable configuration files as `database.yml`, `secrets.yml`, etc. and adding sample ones would be great. So then each developer can set its own environment or use the sample files to set it up (as default).

rewritten:
Closing this, we will have a normal upload of avatars.

PR #181, opened 2016-01-06
New Tagging system 

markets:
Idea: 
- improve the current tagging system usability and behaviour.
- :scissors: Angular related code  
- :scissors: Tags Manager library ~> reuse Select2 (already in use)
- add/enhance related tests


app/assets/javascripts/tags.js.coffee
```diff
@@ -1,17 +1,34 @@
-$(document).on "click", ".switch_offer", (event) ->
-  $.ajax '/tags/offers',
-    type: 'GET'
-    dataType: 'html'
-    error: (jqXHR, textStatus, errorThrown) ->
-        $('.alpha_tag_list').html "AJAX Error: #{textStatus}"
+$ ->
+  $(".switch_offer-js").on "click", ->
```
rewritten: Even if in `$(document).ready()` callback, it's better to use the extended `$(document).on("event", "selector", callback)` so new elements created programmatically can be listened for events. Otherwise only elements that are matched by the initial selector are listened for events.

app/controllers/tags_controller.rb
```diff
@@ -1,60 +1,33 @@
 class TagsController < ApplicationController
-  respond_to :json, :html, :js
-
   def index
     @posts = Post.by_organization(current_organization)
     @all_tags = @posts.find_like_tag(params[:term])
-    respond_with @all_tags
+
+    render json: @all_tags
```
rewritten: :heart:

rewritten:
Nice!!!

app/controllers/tags_controller.rb
markets: La idea es ir quitando `respond_with` (ya quedan pocos) para eliminar `responders` como dependencia.

app/assets/javascripts/tags.js.coffee
markets: Yes, you're right, but in this case I considered that more elements like those (a switches for a very specific view) wouldn't be created programmatically :smile:

app/controllers/tags_controller.rb
rewritten: Perfecto! Además los controllers de API deberían estar separados de los
visuales (remitiendo a un repository común para saber qué records
devolver). Estoy leyendo cosas muy interesantes que me ha pasado Enrico
sobre arquitectura, y la verdad es que separar este tipo de cosas ayuda
mucho en la simplicidad de la aplicación. http://cbra.info

app/assets/javascripts/tags.js.coffee
rewritten: Turbolinks? PJAX? All these are programmatical. And I intend to have Turbolinks 3 in place as soon as we can touch that layer.

markets: Yep :), aunque Turbolinks 3 probablemente nunca será lanazado (https://github.com/rails/turbolinks/issues/628), en su lugar, llega Turbolinks 5 (https://github.com/basecamp/turbolinks/tree/v5)...

rewritten: :cry: sin hooks de replace/append.... que pena.

PR #183, opened 2016-01-18
Drop responders runtime dependency

markets:


markets:
@rewritten what do you think?

rewritten:
:heart:

PR #187, opened 2016-01-25
Dockerize project

d2bit:
That's my environment set up with docker.


rewritten:
Is it working on linux or on OS X too?

rewritten:
That's really cool. Is it like several machines which are available to each other through the 'link' stanza on the compose file? (I'm new to docker...)

Can it be automated and integrated with the ansible scripts for local development?

enricostano:
Nice, remember that we just need a box (docker, LXC, VM, bare metal, PC, VPS, etc) that can be the target of our provisioning process. And that we want to re-use provisioning scripts (Ansbile, etc) between production and development as much as possible.

d2bit:
yeah, it should work over OSX, as docker can run. (with some kind of vbox or so)
I really hate osx :trollface:

enricostano:
For Linux I can offer the LXC setup that we use for Katuma. You don't even need docker.

d2bit:
@enricostano I thought the other PR was about it. I've seen docker containers could be provisioned through ansible, but I didn't touch any ansible (not yet) But seems a really interesting idea

enricostano:
Basic concept shown here in my last Ansible talk https://github.com/enricostano/ansible-lxc-demo

d2bit:
:clap: :clap: I'll take a look later

rewritten:
Ok so with this docker-compose.yml file I can set up all the images, let's see if they run unde xhyve.

rewritten:
Status of this?

d2bit:
In my opinion this could be closed as docker-compose won't be used..

rewritten:
Ok thanks for the effort anyway @d2bit

PR #189, opened 2016-03-18
Fix spec for offers controller search

rewritten:


rewritten:
:green_heart: :green_heart: :green_heart:

spec/controllers/offers_controller_spec.rb
```diff
@@ -60,7 +60,7 @@
       login(another_member.user)
 
       get "index", q: offer.title.split(/\s/).first
-      expect(assigns(:offers)).to eq([offer])
+      expect(assigns(:offers).to_a).to eq([offer])
```
enricostano: what returns `assigns(:offers)`? why do we need the `.to_a`?

rewritten: It returns a proxy from ElasticSearch to a database recordset. It delegates almost everything to the underlying recordset, like `:[]`, `:to_a`, `:each` etc. But it is **not** an array itself.

rewritten: I'll add a comment.

enricostano:
:+1: :clap:

PR #190, opened 2016-03-18
Untrack configuration files, add example files

rewritten:
Remove database and env files from main repository, adding sane examples for forks etc.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@ SECRET_TOKEN=change_me
 # mail configuration:
 MAIL_LINK_HOST=localhost:3000
 MAILER_SENDER=me@example.com
-SMTP_ADDRESS=smtpsmpt_host
+SMTP_ADDRESS=smtp_host
 SMTP_PORT=587
 SMTP_USER_NAME=smtp_username
 SMTP_PASSWORD=smtp_password
--- a/.env.example
+++ b/.env.example
@@ -8,15 +8,23 @@
 # Your local .env file is added to .gitignore and
 # wont be uploaded to the repo.
 
+SECRET_TOKEN=change_me
 
-# database setup:
-DATABASE_ADAPTER=postgres
-DATABASE_USER=postgres
-DATABASE_PASSWORD=postgres
-DATABASE_HOST=localhost
-
-# host part of the url for mail links:
+# mail configuration:
 MAIL_LINK_HOST=localhost:3000
+MAILER_SENDER=me@example.com
+SMTP_ADDRESS=smtpsmpt_host
+SMTP_PORT=587
+SMTP_USER_NAME=smtp_username
+SMTP_PASSWORD=smtp_password
+SMTP_AUTHENTICATION=:plain
+SMTP_ENABLE_STARTTLS_AUTO=true
 
-# a list of emails for superadmin users
+# a space-separated list of emails for superadmin users
 ADMINS="superadmin@example.com"
+
+# default: set to something else if needed.
+# ELASTICSEARCH_URL=localhost
+
+
+# ROLLBAR_ACCESS_TOKEN=rollbar_access_token
--- a/.env.example
+++ b/.env.example
@@ -26,5 +26,5 @@ ADMINS="superadmin@example.com"
 # default: set to something else if needed.
 # ELASTICSEARCH_URL=localhost
 
-
+# Set up your own access to rollbar if needed
 # ROLLBAR_ACCESS_TOKEN=rollbar_access_token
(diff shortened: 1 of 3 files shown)


rewritten:
cc: @d2bit

rewritten:
(it will be green once #189 is merged.)

.env.example
```diff
@@ -8,15 +8,23 @@
...
+SMTP_ENABLE_STARTTLS_AUTO=true
 
-# a list of emails for superadmin users
+# a space-separated list of emails for superadmin users
 ADMINS="superadmin@example.com"
+
+# default: set to something else if needed.
+# ELASTICSEARCH_URL=localhost
+
+
```
enricostano: white space

.gitignore
```diff
@@ -35,6 +35,7 @@ timeoverflow_development
 timeoverflow_test
 
 .env
+config/database.yml
```
enricostano: can we leave `config/database.yml` tracked and use environment variables for `production` environment?

I say so to avoid the `copy to database.yml` thing

rewritten: Everyone has its own local database. Typically, pg on OS X has different defaults for pg on linux, and pl may have their own db server in a raspberry π.

enricostano: I never found this to be a problem, I would leave this step as easy as possible and with a standard for `development` and `test` #RailsDeTodaLaVida

Then hackish people can override that option as they wish. In this way the development environment thing will be easier too.

markets: Something like this?

```
defaults: &defaults
  adapter: postgresql
  host: <%= ENV.fetch('DATABASE_HOST', 'localhost') %>
  username: <%= ENV.fetch('DATABASE_USERNAME', 'root') %>
  encoding: utf8

development:
  <<: *defaults
  database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_development') %>

test:
  <<: *defaults
  database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_test') %>

production:
  <<: *defaults
  database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_production') %>
```

enricostano: yep, I would add `password: <%= ENV['DB_PASSWORD'] %>` to `defaults` too

d2bit: I use to have an untracked `database.yml` in my environment, so I can change it as my wish. 

On the other hand, people who can't copy a database.yml.sample to database.yml maybe should learn to do it..
Just uncomment these lines and :tada: https://github.com/coopdevs/timeoverflow/blob/master/bin/setup#L15-L18

I've been looking for the initial setup in the `Getting started guide` and seems we are not using the `bin/setup` script..  maybe it will make setup a little easier.

rewritten: It would work for me with the following (not `root`, just plain peer trust)

```yml
defaults: &defaults
  adapter: postgresql
  host: <%= ENV.fetch('DB_HOST', 'localhost') %>
  username: <%= ENV.fetch('DB_USERNAME', 'null') %>
  password: <%= ENV.fetch('DB_PASSWORD', 'null') %>
  encoding: utf8
```

enricostano: We should decide what we want to use and try to stick to it, because everybody can come up with a slightly different flavor. Furthermore we should use what we decide to use for the development environment for TO, that's almost complete BTW. The trade off is between what we want in production and in the official development environment, and make that decision as customizable as possible using environment variables.

rewritten: And that's the reason of my proposal. Having peer ident in a local postgres setup seems like a very reasonable default. And this requires having 'null' as username and password.

As always, more expert people can use anything from the full DATABASE_URL to the specific DB_HOST/DB_USERNAME/DB_PASSWORD env variables.

enricostano: seems fair to me

rewritten: See #192

rewritten: We can also explain the --assume-unchanged git flag, that would allow to stop tracking changes to a tracked files.

rewritten:
So the only thing left here is the update of `.env.example` which I think is worth (btw I'll update the wiki for that). But maybe in a separate and cleaner PR. Opinions?

rewritten:
Wiki updated

.env.example
```diff
@@ -8,15 +8,23 @@
 # Your local .env file is added to .gitignore and
 # wont be uploaded to the repo.
 
+SECRET_TOKEN=change_me
```
enricostano: are we using this? I mean in production we populate it during provisioning, right?

rewritten: indeed.

sauloperez:
@enricostano agree on closing this? See my comment above.

PR #191, opened 2016-03-19
Speed up TravisCI database setup

d2bit:
Use db:setup db:migrate instead of just db:migrate
Don't install production & development gems
Configure db from database.yml

Let's see if this runs faster than before..
BTW.. that's how I use to work with the `database.yml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,11 @@
 language: ruby
-
+sudo: false
+cache: bundler
+bundler_args: '--without production development'
 rvm:
   - 2.2.2
-
 env: DATABASE_URL=postgres://postgres@localhost/timeoverflow_test
-
 before_script:
-  - psql -c 'create database timeoverflow_test;' -U postgres
-  - RAILS_ENV=test bundle exec rake db:migrate
-
-sudo: false
-
-cache: bundler
-
+  - bundle exec rake db:setup
 services:
   - elasticsearch


.travis.yml
```diff
@@ -1,17 +1,14 @@
 language: ruby
+sudo: false
+cache: bundler
+bundler_args: '--without production development'
 
 rvm:
   - 2.2.2
 
-env: DATABASE_URL=postgres://postgres@localhost/timeoverflow_test
```
rewritten: Why not the `DATABASE_URL` ? Besides making 12-factors more commonplace, in my opinion this is the biggest contribution of Heroku to the rails app concept.I would love to see the same for SMTP in fact.

.travis.yml
```diff
@@ -1,17 +1,14 @@
...
-
 before_script:
-  - psql -c 'create database timeoverflow_test;' -U postgres
-  - RAILS_ENV=test bundle exec rake db:migrate
-
-sudo: false
-
-cache: bundler
+  - cp config/database.example.yml config/database.yml
+  - bundle exec rake db:setup db:migrate
```
rewritten: :+1:

.travis.yml
d2bit: What I try with this is to have the CI configured in the same place as the other. So that we avoid introducing any differences between them.

rewritten: Well, production goes with DATABASE_URL, so it makes sense that CI goes with the same.

enricostano:
I really see the `database.yml` thing an overcomplication, but if there is consensus... I'm OK with that.

markets:
I'm with @enricostano on this. Yes, rename a file is an easy step, but do nothing is even easier, so why not  ¯\_(ツ)_/¯ ?

d2bit:
Okay then. I just made a suggestion :stuck_out_tongue_winking_eye:

rewritten:
:+1: for me-

enricostano:
So, we need more votes here! @sauloperez `#mojate`

sauloperez:
I agree with @enricostano and @markets . Something like https://github.com/coopdevs/timeoverflow/pull/190#discussion_r56692395 seems the easiest and most flexible solution. In the end, if you need to customize any value, you just need to define the proper env vars.

rewritten:
@d2bit can we just change the `db:setup` + `db:migrate` and leave the debate about `database.yml` for #190 ?

I will be more than happy to merge this with spot-on changes.

Rebase (one-commit) and force push if you wish.

markets:
@d2bit :+1: 

Just a comment, why not keep a white line between each section? Or just remove all those blank lines... Consistency is good for :eyes: :smile:

.travis.yml
```diff
@@ -6,12 +9,7 @@ rvm:
 env: DATABASE_URL=postgres://postgres@localhost/timeoverflow_test
 
 before_script:
-  - psql -c 'create database timeoverflow_test;' -U postgres
-  - RAILS_ENV=test bundle exec rake db:migrate
-
-sudo: false
-
-cache: bundler
+  - bundle exec rake db:setup db:migrate
```
enricostano: > The rails db:setup task will create the database, load the schema and initialize it with the seed data

do we really need `db:migrate` here?

d2bit: Not really, but.. could we forget to run migrations or add changes to
schema.rb?
I think it could be useful.. but not mandatory.

enricostano: Every time we add a new migration, in the same PR we add the modified schema. We must ensure this. We generally enforce this running a `bad practices` suite. I would rely on our capability to don't break the schema, even more.. if we break the schema is better to know that in travis testing a PR than directly in production.

What do you think?

d2bit: yeah you're absolutely right! :+1:
I'll amend it  tomorrow

rewritten:
So we need a PR with a "schema is up to date" test. How can it be implemented?

enricostano:
> So we need a PR with a "schema is up to date" test. How can it be implemented?

The "tool" would be really simple, like just checking if the last migration id is reflected in the schema file. Travis would check every PR.

**But**, I wouldn't say that we need that tool right now. We can just start with code review, we have other priorities IMO.

d2bit:
removed `db:migrate` call and _'inconsistent'_ spaces in `travis.yml`

d2bit:
BTW, I think Rails will complain if there are migrations pending to run...
```
Migrations are pending. To resolve this issue, run:
        bin/rake db:migrate RAILS_ENV=test
```

So it's covered

rewritten:
Yep. you are right, I supposed it was a warning message but it's actually an error for the rake/spec/whatever script, so CI would give a :red_circle:

:+1:  Thanks @d2bit

enricostano:
:clap: :smiley:

PR #192, opened 2016-03-22
Reasonable defaults for database.yml

rewritten:
Trying to settle #190


enricostano:
What about `production` environment?

rewritten:
As already said somewhere else, production is configured with DATABASE_URL. We should document it but not here.

enricostano:
What about adding something like:
```
production:
  url: <%= ENV.fetch('DATABASE_URL', 'Please set DATABASE_URL environment variable') %>
```
*the fallback string can be whatevarrr :smiley:

rewritten:
:+1:

enricostano:
@d2bit @markets @sauloperez @rewritten thumbs? :smiley:

markets:
:+1: :ok_hand: :tada: 

We can probably port to this branch the changes related to `.env.example`, made by @rewritten in #190, what do you think?

d2bit:
:+1: but I see strange the `'null'` :trollface:

rewritten:
@markets http://youtu.be/G2gTFBhQ7Ko
@d2bit I don't love it either, but it's erb fault, not yaml. `<% if ENV.key? "DATABASE_PASSWORD" %>password: <%= ENV["DATABASE_PASSWORD"] %><% end %>` is **much** worse.

rewritten:
I will refactor #190

markets:
Yep, 100% agree, but as I have seen the same intention in #190... I supposed this should move on a similar direction :stuck_out_tongue:

enricostano:
@d2bit @rewritten instead on `'null'`, what about:
```
  username: <%= ENV.fetch('DATABASE_USER', nil) %>
  password: <%= ENV.fetch('DATABASE_PASSWORD', nil) %>
```
or those `nil` will break Postgres in some way?

rewritten:
As discussed in chat, having `nil` in the ERB tag would lead to invalid YAML. One possible option is to output something (falsey but not blank) for which `YAML.parse` is a left compositional inverse of `&:to_s` (which is a really fancy and absolutely informal way to say that `x == YAML.parse(x.to_s)`) and moreover where Rails likes the value and understands that there is "no password".

In any case this is becoming complex, can we just :+1:  it at this stage and delay any further evolution of this single file for later?

PR #193, opened 2016-03-23
Session, cookies and security

enricostano:
### What is this?
We found out that people tend to leave their TO account logged in while using public computers. This is a security concern, of course.

We use cookies to store sessions, this is not the best solution since cookies can be stolen and we should switch to some database store for sessions (Memcache, Redis, etc). But that's out of the scope of this PR.

With this PR we're addressing:

1. Update Devise gem to get some important security update
2. Adopt the new Rails `secrets.yml` and remove a deprecation warning, more info [here](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#config-secrets-yml)
3. Add an expiration to the session cookie

### How to test it
1. Test that you must log in again after 1 hour you stopped using TO
2. Test that if you check the option `Remember me` during login it shouldn't ask to log in again during 2 weeks


rewritten:
Maybe it can be tested using Time.travel_to (for the session duration) and abusing the test client so non-persistent cookies are deleted (to check for "remember me" tokens)

Gemfile.lock
```diff
@@ -394,7 +394,7 @@ DEPENDENCIES
   coffee-rails
   dalli
   database_cleaner
-  devise
+  devise (= 3.5.6)
```
rewritten: I prefer soft dependencies (`~> 3.5.6`) if possible.

rewritten:
Are the three changes really related? Devise update, secret token configuration and session expiry? I don't see it.

rewritten:
Final question: given that the session expiration is counted from the last request, isn't 1 hour a little too long? That is 1 hour of leaving the computer unattended.

Gemfile.lock
enricostano: Well, we'll need another consensus then. I've been bitten by soft dependencies so much that I don't understand why it's not the Gemfile default in official examples. You must ensure which version you're using and yes, I don't trust maintainers.

enricostano:
> Are the three changes really related? Devise update, secret token configuration and session expiry? I don't see it.

Nope, I can separate this PR in 3 PRs if needed.

enricostano:
> Final question: given that the session expiration is counted from the last request, isn't 1 hour a little too long? That is 1 hour of leaving the computer unattended. 

I put `1.hour` but we need to understand what's the right trade off between security and usability. Only time and people complains will tell the truth.

@sseerrggii any idea about this point?

Gemfile.lock
rewritten: Ok, let's go the hard version way, no problem. In any case `bundle outdated` would show us whether there is any update pending.

rewritten:
so I think that two thirds of this can be merged right away. I start to see the point of very small PRs.

enricostano:
> so I think that two thirds of this can be merged right away. I start to see the point of very small PRs.

Small PRs es bien :smile: 

Let's wait for some input on the expiration duration and then we'll merge this.

sseerrggii:
I think 1 hour is a good solution :+1:

rewritten:
Ok merging? or do we want to try and make tests for this? (My opinion is that testing this is part of the framework and we shouldn't test the framework)

fcsonline:
:+1:

enricostano:
OMG we have @fcsonline 's blessing! :scream:

fcsonline:
hackers!!!

enricostano:
BTW, we should change the `SECRET_TOKEN` in production to invalidate all the old cookies. Will do in provisioning repo.

PR #194, opened 2016-03-24
Rails complains when the `database.url` key is not an url

rewritten:
This is the output of `rake db:migrate:status` (for instance) whatever the environment is:
```
rake aborted!
URI::InvalidURIError: bad URI(is not URI?): Set DATABASE_URL environment variable
/.../config/environment.rb:5:in `<top (required)>'
Tasks: TOP => db:migrate:status => environment
(See full trace by running task with --trace)
```

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -14,4 +14,5 @@ test:
   database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_test') %>
 
 production:
-  url: <%= ENV.fetch('DATABASE_URL', 'Set DATABASE_URL environment variable') %>
+  # Set DATABASE_URL environment variable
+  url: <%= ENV.fetch('DATABASE_URL', 'null') %>
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,7 +1,7 @@
 # If you want to change this file, please keep the changes in your working
 # copy by using
 #
-#     git update-index --assume-unchanged config/database.yml
+#     git update-index --skip-worktree config/database.yml
 #
 
 defaults: &defaults
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,8 +1,8 @@
 defaults: &defaults
   adapter: postgresql
-  host: <%= ENV.fetch('DATABASE_HOST', 'localhost') %>
-  username: <%= ENV.fetch('DATABASE_USER', 'null') %>
-  password: <%= ENV.fetch('DATABASE_PASSWORD', 'null') %>
+  host: <%= ENV['DATABASE_HOST'] %>  # default is localhost
+  username: <%= ENV['DATABASE_USER'] %>  # default is null
+  password: <%= ENV['DATABASE_PASSWORD'] %>  # default is null
   encoding: utf8
 
 development:
@@ -15,4 +15,4 @@ test:
 
 production:
   # Set DATABASE_URL environment variable
-  url: <%= ENV.fetch('DATABASE_URL', 'null') %>
+  url: <%= ENV['DATABASE_URL'] %>
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,6 +3,12 @@
 #
 #     git update-index --skip-worktree config/database.yml
 #
+# or just use DATABASE_URL, in which case Rails will happily skip the whole
+# file.
+#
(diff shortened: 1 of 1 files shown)


rewritten:
@enricostano urgent.

config/database.yml
```diff
@@ -14,4 +14,5 @@ test:
   database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_test') %>
 
 production:
-  url: <%= ENV.fetch('DATABASE_URL', 'Set DATABASE_URL environment variable') %>
+  # Set DATABASE_URL environment variable
+  url: <%= ENV.fetch('DATABASE_URL', 'null') %>
```
d2bit: http://guides.rubyonrails.org/configuring.html#connection-preference
>If you have an empty config/database.yml file but your ENV['DATABASE_URL'] is present, then Rails will connect to the database via your environment variable

So that line could/should be removed

rewritten: If we remove only that line, it is invalid YAML format. so either we skip the production section (and in fact in actual production there is `ENV['DATABASE_URL']` exactly for that reason) or we leave a section just for completeness, there was already a conversation about that, and we ended up adding that section for the [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

d2bit: `production` key could be removed and Rails also will read the DATABASE_URL for that environment..

other option could be to apply defaults as in the other environments as the DATABASE_URL will always be merged
```yaml
production:
  <<: *defaults
  database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow') %>
```

rewritten:
Ok so the database.yml file can have actually empty values for some keys (for sure not the environment keys, which are supposed to be hashes, but at least the ones for configuration). So instead of the ugly `ENV.fetch(K, 'null')` it's much better to have just the normal ruby hash fallback to nil, and let ERB write out an empty string, which will be swallowed by rails :)

enricostano:
let's test this in all 3 environments and :+1:

rewritten:
already tested (by changing the RAILS_ENV locally)

rewritten:
But... http://fallengamer.livejournal.com/93321.html so `--assume-unchanged` is not robust. `--skip-worktree` is the correct flag for this use case.

rewritten:
@d2bit I think this (with the `--skip-worktree` bit) covers your use case too, without adding the "copy the example file" step. What do you think?

config/database.yml
```diff
@@ -1,8 +1,14 @@
+# If you want to change this file, please keep the changes in your working
+# copy by using
+#
+#     git update-index --skip-worktree config/database.yml
```
enricostano: can we add a suggestion to use environment variables and as a last option the update-index thing?

enricostano: BTW, maybe we should just create a wiki or doc in markdown somewhere like `How to override default DB configuration` instead of putting all this stuff in the code

rewritten: Why do you prefer env to --skip-worktree? This is exactly the use case for this flag, as you can read in the linked page above: Have developers change a fil to suit their local needs, but keep the original file forever in the repository. The skipped file will never be comitted, but if a remote change creates a conflit, then the user is asked to resolve the conflict locally, and the new version is still skipped. It's a win-win.

And, if you know that rails loves the DATABASE_URL env variable, you are done and you don't need any more instructions.

About having it in the wiki, I disagree for two reasons:

1. people don't read
2. people might fork the repo, and not the wiki
3. this is really specific to the database.yml file, and people is used to look in there to configure their local settings
4. people can't count.

enricostano: > people don't read
Adding a comment in the code with a link to the documentation could help

> people might fork the repo, and not the wiki
I agree, but we're just trying to be nice adding documentation on how to customize the default setup

enricostano: > And, if you know that rails loves the DATABASE_URL env variable, you are done and you don't need any more instructions.

So why we're having this discussion? Sorry if I'm missing anything :pray:

enricostano:
:+1:

config/database.yml
rewritten: @enricostano I'm trying to make both you and @d2bit happy. @d2bit wants a local untracked file (achieved with --skip-worktree), you want a complete database.yml (you have it). I want a pretty much empty production section and no user/password in development, I have it.

PR #195, opened 2016-03-24
Test that database schema is up to date

rewritten:


d2bit:
I think Rails do it by default.
I've tried to add a migration, not running it, and then rspec complains about having pending migrations
Can you try in your environment? :confused: 

I've said just before you merged the PR.
https://github.com/coopdevs/timeoverflow/pull/191#issuecomment-200725029

rewritten:
@d2bit completely right, it's not needed.

PR #203, opened 2016-04-24
Add date at inquiries list

sseerrggii:


rewritten:
Inspection/reflection in a view, there should be a better way. Why not adding the date for offers too?

sseerrggii:
Because an inquiry we may want to know how old it is, for example if someone asks for help with the move and has more than 3 months worth no longer interested. Instead the offers are more timeless. This improvement serves just to see how old the inquiries are in the list

rewritten:
Ok then let's hide the date by some other means for "timeless" objects. Like in css?

```css
.offer.creation-time {
  display: none;
}
```

@markets @enricostano @naoisegolden any comment, any other solution that does not involve introspection in a view?

enricostano:
I agree with @rewritten about having `post.type == "Inquiry"` in the view (although that check in general is a bad practice).

The CSS solution doesn't seem the best one neither IMHO.

Maybe we just want to create another partial like `_inquiries_posts.html.haml`. DRY principle doesn't always apply.

rewritten:
Yes, a different partial may be a good idea.

markets:
If we're going to split into 2 views, what about moving the code to the respective detail page (instead of nested partials):

- offer: https://github.com/coopdevs/timeoverflow/blob/master/app/views/offers/show.html.haml
- inquiry:https://github.com/coopdevs/timeoverflow/blob/master/app/views/inquiries/show.html.haml

And ✂️  `app/views/shared/_posts.html.haml`.

enricostano:
@markets yes, could be an option since we'll not reuse them for now... but I really hate BIG files :joy:

markets:
Ah yes @enricostano, BIG files are not nice, but nested partials or views composed of lots of partials neither :) just my 2 cents...

By the way, it'll be a ~35 lines file.

sseerrggii:
Al final para no complicarlo y hablándolo con @enricostano hemos decidido poner la fecha tanto para listados de ofertas como de demandas

enricostano:
Consistency es bien. Y te ahorra over-complication.

👍

PR #205, opened 2016-04-27
green and red in hours positive or negative

sseerrggii:
Green or red on movements and balance depending on whether positive or negative, so it is easier to see if they are inputs or outputs of hours

### DIFF
--- a/app/views/shared/_movements.html.haml
+++ b/app/views/shared/_movements.html.haml
@@ -38,6 +38,8 @@
                 - op = { method: :put, data: { confirm: t(".delete_reason") } }
                 = link_to(delete_reason_transfer_path(mv.transfer), op) do
                   %span.glyphicon.glyphicon-trash
-            %td= seconds_to_hm(mv.amount)
+            %td
+              %strong{ class: green_red(mv.amount) }
+                = seconds_to_hm(mv.amount)
       .paginate-align
         = paginate @movements
--- a/app/views/users/show.html.haml
+++ b/app/views/users/show.html.haml
@@ -103,8 +103,8 @@
             = t(".user_no")
           = @member.member_uid || mdash
           %br
-          %strong
+          %strong.lead{ class: green_red(@member.account&.balance) }
             = t(".balance")
-          = seconds_to_hm(@member.account.try(:balance) || mdash)
+            = seconds_to_hm(@member.account.try(:balance) || mdash)
 
 = render "shared/movements"
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,6 +1,10 @@
 require "date"
 
 module ApplicationHelper
+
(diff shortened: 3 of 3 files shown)


app/views/shared/_movements.html.haml
```diff
@@ -38,6 +38,9 @@
                 - op = { method: :put, data: { confirm: t(".delete_reason") } }
                 = link_to(delete_reason_transfer_path(mv.transfer), op) do
                   %span.glyphicon.glyphicon-trash
-            %td= seconds_to_hm(mv.amount)
+            %td
+              - positive = mv.amount.positive?
+              %strong{ :class => (positive ? "text-success" : "text-danger") }
```
enricostano: Creo que puedes directamente poner:
```
%strong{ :class => mv.amount.positive? ? "text-success" : "text-danger" }
```
y ahorrarte la asignacion de la variable `positive`

app/views/users/show.html.haml
```diff
@@ -103,8 +103,9 @@
             = t(".user_no")
           = @member.member_uid || mdash
           %br
-          %strong
+          - positive = @member.account.try(:balance).positive?
+          %strong.lead{ :class => (positive ? "text-success" : "text-danger") }
```
enricostano: lo mismo

app/views/shared/_movements.html.haml
sseerrggii: Creo que si no se pasa de tamaño de linea

enricostano: `%strong{ class: mv.amount.positive? ? "text-success" : "text-danger" }`
te he quitado 3 caracteres mas :joy:

enricostano: BTW... lo has probado? https://travis-ci.org/coopdevs/timeoverflow/builds/126176717#L429

esto falla por lo de `positive?`

sseerrggii: A mi me va

enricostano: `positive?` is available in ruby `2.3.0` but not in `2.2.2` which is the version that we use in travis.

https://github.com/coopdevs/timeoverflow/pull/206 fix this

grande @sseerrggii :clap:

rewritten:
In a helper:

```ruby
TEXT_SUCCESS = 'text-success'.freeze
TEXT_DANGER = 'text-danger'.freeze

def green_red(amount)
  case amount <=> 0
  when -1 then TEXT_DANGER
  when 0 then nil
  when 1 then TEXT_SUCCESS
  end
end
```

And in the view:

```haml
%strong.lead{class: green_red(@member.account&.balance)
```

Note the use of the safe dot (`&.`) instead of try, again as a nice gift from Matz in ruby 2.3

enricostano:
👍

PR #212, opened 2016-05-13
Rubocop: enable much saner defaults

rewritten:


Gemfile
```diff
@@ -46,28 +46,28 @@ gem 'select2-rails'
...
 end
 
 group :development, :test do
-  gem "rspec-rails", '~> 3.4.0'
-  gem "capybara", '~> 2.4.4'
-  gem "byebug"
+  gem 'rspec-rails', '~> 3.4.0'
+  gem 'capybara', '~> 2.4.4'
+  gem 'byebug'
+  gem 'rubocop', '~> 0.40'
```
d2bit: Do you plan to add a rubocop step to the test suite? :confused:

rewritten: @d2bit  HoundCI does it, just for things that change in each PR. Having rubocop check the whole system is a big rewrite, but at least all new code will be covered.

d2bit: Aham, so I guess that adding the gem in testing environment.. HoundCI will use that exact version :+1:
Didn't know about that..

rewritten: Not sure about that, actually I doubt it, but I think it makes sense to have it on Travis at some point.

enricostano:
> agree on closing this @enricostano ? Things have changed considerably regarding Rubocop since then and I believe it deserves to be worked on again.

Yep. Please anyone ping us if you want to work on this again, but in another PR I guess. Thanks everybody!

PR #263, opened 2017-09-08
Search only active posts

sauloperez:
I'm trying to solve https://github.com/coopdevs/timeoverflow/issues/262 but I found an issue for which I don't have an explanation yet.

As you see below, I had to add the `active` field in the ES index for posts. We want to find active posts only as we do with the DB query for `PostsController#index` normally.

Once I recreate the index I get:

```shell
$ curl localhost:9200/offers/_mappings?pretty=1
{
  "offers" : {
    "mappings" : {
      "offer" : {
        "properties" : {
          "active" : {
            "type" : "boolean"
          },
          "description" : {
            "type" : "string",
            "analyzer" : "normal"
          },
          "organization_id" : {
            "type" : "integer"
          },
          "tags" : {
            "type" : "string"
          },
          "title" : {
            "type" : "string",
            "analyzer" : "normal"
          }
        }
      }
    }
  }
}
```

Which is what I expect. However when I query all the documents in the index:

```shell
$ curl -X POST localhost:9200/offers/_search?pretty=1 -d '{"query": { "match_all": {} } } }' -H "Content-Type: application/json"
...
}, {
      "_index" : "offers",
      "_type" : "offer",
      "_id" : "6",
      "_score" : 1.0,
      "_source" : {
        "title" : "Pequeñas reparaciones de casa",
        "description" : "  Llamame y miramos que necesitas, si veo que puedo ayudarte nos ponemos a ello\n",
        "tags" : [ "casa", "manitas" ],
        "organization_id" : 1
      }
    }, {
...
```

`_source` contains all fields but `active` :thinking: As a result no query I tried specifying `active: true` worked.

The hypothesis I had was that perhaps `active` is a reserved word but in that case I'd expect seeing a failure on Elasticsearch's logs when import the records to it or when parsing the query. I can't seem to find a list of reserved words either...

@enricostano feel free to this PR over if you want while I'm off. This is all I could find.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -6,28 +6,32 @@ class PostsController <  ApplicationController
   def index
     if (query = params[:q]).present?
       # match query term on fields
-      must = [ { multi_match: {
-          query: query.to_s,
-          type: "phrase_prefix",
-          fields: ["title^2", "description", "tags^2"]
-        } } ]
+      must = [
+        {
+          multi_match: {
+            query: query.to_s,
+            type: "phrase_prefix",
+            fields: ["title^2", "description", "tags^2"]
+          }
+        }
+      ]
+
       if current_organization.present?
         # filter by organization
         must << { term: { organization_id: { value: current_organization.id } } }
       end
-      posts = model.__elasticsearch__.search(
-        query: {
-          bool: {
-            must: must
-          }
-        }
-      ).page(params[:page]).per(25).records
+
+      posts = model.__elasticsearch__.search( query: { bool: { must: must } })
+      posts = posts.page(params[:page]).per(25).records
     else
       posts = model.active.of_active_members
+
       if current_organization.present?
         posts = posts.merge(current_organization.posts)
       end
-      posts = apply_scopes(posts).page(params[:page]).per(25)
+
+      posts = apply_scopes(posts)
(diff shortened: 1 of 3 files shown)


app/controllers/posts_controller.rb
```diff
@@ -6,28 +6,32 @@ class PostsController <  ApplicationController
...
     else
       posts = model.active.of_active_members
+
       if current_organization.present?
         posts = posts.merge(current_organization.posts)
       end
-      posts = apply_scopes(posts).page(params[:page]).per(25)
+
+      posts = apply_scopes(posts)
+      posts = posts.page(params[:page]).per(25)
```
sauloperez: I think you @enricostano see what I already did here. `Turn difference into sameness` 😆 . If you look at it, we're doing exactly the same (both branches of the conditional look almost equal), but in two different data stores.

No need to build a castle, but we can abstract things a bit and remove duplication.

sseerrggii:
we can close? @sauloperez

sauloperez:
If #262 never happened again yep, I would close this one.

sauloperez:
This won't be needed once we replace ES with Potsgres.

enricostano:
> This won't be needed once we replace ES with Potsgres.

Is this already a fact @sauloperez ?

sauloperez:
Nope, the initial investigation still needs to happen.

PR #273, opened 2017-10-18
Update deps

knoopx:
It also closes #284

PR #283, opened 2018-01-11
Upgrade to rails 4.2.7.1

sauloperez:
This will solve the most severe of the security vulnerabilities highlighted by GitHub. They are related to ActionView and ActiveRecord so we would be better off upgrading. You can see the full list of changes in https://github.com/rails/rails/compare/v4.2.5.2...v4.2.7.1. Besides, as we all know it's a good practice to keep up to date ASAP.

The upgrade should be harmless as the changes are very limited. As soon as this passes CI, I'd give a general testing round and merge if we're happy with it.


sauloperez:
I had to upgrade RSpec Rails as well in order to get rid of this well known failure https://github.com/ruby/rake/issues/116.

sauloperez:
#273 goes further and upgrades Rails to 4.2.10 among other dependencies. I suggest we merge this one first so that then #273 it's a smaller change `#conservative`. Thoughts? cc @enricostano @sseerrggii

sseerrggii:
Tested :+1:

PR #291, opened 2018-01-18
Improve update balance

sauloperez:
This adds specs to `Account#update_balance`, docs and refactors the method a bit. It extracts compound conditionals, adds some doc and emphasizes problems that arise from the current implementation such as the fact that max_allowed_balance it's either Integer or Float.


sauloperez:
thanks for the quick review @enricostano!

PR #272, opened 2017-10-18
Refactor Users#index to paginate and filter on the backend. Closes #266

knoopx:
* Results are now paginated using kaminari on the server side
* Filtering and sorting is performed now on the server side using ransack, which was already part of the bundle (activeadmin depends on it) and because, in my opinion, elasticsearch was kind of overkill for this specific case
* Removed angular completely (couldn't find any other usage within the app)

PR #292, opened 2018-01-18
Remove TODO. This method is used in users list

sauloperez:
So, we can't eliminate the method and it's not that important to refactor. First, it should have specs... In any case, the problems that we have right now are not related to this method.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -1,5 +1,4 @@
 module UsersHelper
-  # TODO refactor or eliminate - poosibly the second.
   def users_as_json
     @users = (admin? || superadmin?) ? @users : @users.actives
     @users.map do |user|


enricostano:
@knoopx already dealt with it 💪 

https://github.com/coopdevs/timeoverflow/pull/272/files#diff-1f527483fa5f6b74202ae2c2f0c06af0

sauloperez:
🤔  but I found that:

> This method is used in users list

if I remove it the user list is totally empty as we don't populate that data for angular to render it...

enricostano:
You cannot remove it yet, just wait for the other PR.

sauloperez:
Ok, got it

enricostano:
I wouldn't even remove the TODO right now, that would create conflicts in the other PR and slower it even more 🙈

sauloperez:
Yes, sure, no need to spend any more time on it.

PR #264, opened 2017-09-19
Move give time action to transfers controller

sauloperez:
This removes the duplicated action `#give_time` (along with its views) in users and organizations controllers and moves it to the `TransfersController#new`.

user_transfer
organization_transfer

## Missing
- [x] Couple screenshots
- [x] Test and docs for `TransferFactory`
- [x] Update current `:give_time` links for organization

PR #316, opened 2018-02-01
Fix/pay from offer

### LINKED ISSUE #315: Crash on transferring time from the offer's page
Looks like we're missing one parameter from the payment button on the offer page.

It should be https://www.timeoverflow.org/transfers/new?destination_account_id=22&id=6&offer=12597
and from the button, we link to https://www.timeoverflow.org/transfers/new?id=6&offer=12597. This happens when clicking on the top-right green button on an offer's page.

sauloperez:
## What? Why?

Closes #315.

I forgot to provide the destination of the account in the link to transfer, showed on the offer page. This was fetched in `UsersController#give_time` and later passed to the `give_time` view.

When I did the refactor I forgot to move that from https://github.com/coopdevs/timeoverflow/pull/264/files#diff-4e05ad0d64e6100656b63ad1e78f32c5L67 to the appropriate place (the `OffersController#show`), so that we specify the destination in the `link_to`.

## Introducing view specs :tada: 

Since this had to do also with the view, I thought it was the perfect timing to start doing view specs. I've wanted to add them soon to replace the pseudo pattern matching that we ended up doing in some controller specs to ensure stuff in the views.

I figured it'd be faster to get the test I wanted for the view in this way rather than using the current approach in the controller spec. And I was right. Having the capybara RSpec matchers available, such as `have_link` help a lot. You spend less time checking how does the actual returned HTML look like.

### DIFF
--- a/app/controllers/offers_controller.rb
+++ b/app/controllers/offers_controller.rb
@@ -18,4 +18,13 @@ class OffersController < PostsController
       offers[category] = list if list.present?
     end
   end
+
+  def show
+    super
+    @destination_account = @offer
+      .user
+      .members
+      .find_by(organization: current_organization)
+      .account
+  end
 end
--- a/app/controllers/offers_controller.rb
+++ b/app/controllers/offers_controller.rb
@@ -21,10 +21,8 @@ class OffersController < PostsController
 
   def show
     super
-    @destination_account = @offer
-      .user
-      .members
-      .find_by(organization: current_organization)
-      .account
+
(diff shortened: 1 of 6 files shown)


app/controllers/offers_controller.rb
```diff
@@ -18,4 +18,16 @@ def dashboard
       offers[category] = list if list.present?
     end
   end
+
+  def show
+    super
+
+    if current_user
```
enricostano: I'd rather move this check in the view

sseerrggii:
Works fine :+1:

PR #319, opened 2018-02-02
Add unicorn configuration

enricostano:


enricostano:
> We'll stick to the default one?

@sauloperez almost, I modified a bit one of the official examples https://bogomips.org/unicorn.git/tree/examples/unicorn.conf.rb

PR #329, opened 2018-02-23
Added the device_tokens endpoint

mllocs:
In order to send push notifications we need a place to store the device tokens for each users. In this PR:

- I'm adding a new table (`device_tokens`) with a minimal API (just the create action).
- A couple of services and classes to send push notifications.

The future TimeOverflow mobile app will be implemented in this repo: https://github.com/coopdevs/timeoverflow-mobile-app 👶📱

### DIFF
--- a/app/controllers/device_tokens_controller.rb
+++ b/app/controllers/device_tokens_controller.rb
@@ -1,0 +1,19 @@
+class DeviceTokensController < PostsController
+  before_filter :authenticate_user!
+
+  def create
+    @device_token = DeviceToken.all.build device_token_params.merge! user_id: current_user.id
+
+    if @device_token.save
+      render nothing: true, status: :created
+    else
+      render nothing: true, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def device_token_params
+    params.permit(:token)
+  end
+end
--- a/app/controllers/device_tokens_controller.rb
+++ b/app/controllers/device_tokens_controller.rb
@@ -2,7 +2,7 @@ class DeviceTokensController < PostsController
   before_filter :authenticate_user!
 
   def create
-    @device_token = DeviceToken.all.build device_token_params.merge! user_id: current_user.id
+    @device_token = DeviceToken.new device_token_params.merge! user_id: current_user.id
 
     if @device_token.save
       render nothing: true, status: :created
--- a/app/controllers/device_tokens_controller.rb
+++ b/app/controllers/device_tokens_controller.rb
@@ -1,4 +1,4 @@
-class DeviceTokensController < PostsController
+class DeviceTokensController < ApplicationController
   before_filter :authenticate_user!
 
   def create
--- a/app/services/push_notifications/expo_adaptor_service.rb
+++ b/app/services/push_notifications/expo_adaptor_service.rb
@@ -1,0 +1,28 @@
+require 'net/http'
+
+module PushNotifications
+  class ExpoAdaptorService
+    def initialize(notification, user)
+      @notification = notification
+      @user = user
+    end
+
+    def send
+      @user.device_tokens.each do |device_token|
+        # https://docs.expo.io/versions/latest/guides/push-notifications.html
+        uri = URI('https://exp.host/--/api/v2/push/send')
+        Net::HTTP.post_form(uri, post_data(device_token.token))
+      end
+    end
+
+    private
+
+    def post_data(token)
+      {
+        "to" => token,
+        "title" => @notification.title,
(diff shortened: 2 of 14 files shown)


app/controllers/device_tokens_controller.rb
```diff
@@ -0,0 +1,19 @@
+class DeviceTokensController < PostsController
+  before_filter :authenticate_user!
+
+  def create
+    @device_token = DeviceToken.all.build device_token_params.merge! user_id: current_user.id
```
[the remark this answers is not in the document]
sauloperez: It's enough with a `DeviceToken.new(...)`, isn't it?

app/services/push_notifications/expo_adaptor_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+    end
+
+    private
+
+    def post_data(token)
+      {
+        "to" => token,
+        "title" => @notification.title,
+        "body" => @notification.body
```
sauloperez: Please, use attr_readers for the ivars

app/services/push_notifications/post_broadcaster_service.rb
```diff
@@ -0,0 +1,15 @@
+module PushNotifications
+  class PostBroadcasterService
+    def initialize(post)
+      @post = post
+    end
+
+    def broadcast
+      notification = PostNotification.new
+
+      @post.organization.users.each do |user|
```
sauloperez: I would decouple this class from the details of the data model by passing the list of users in the constructor instead. This way if we ever change the data model we don't need to touch the push notifications implementation.

db/migrate/20180221161343_create_device_tokens.rb
```diff
@@ -0,0 +1,12 @@
+class CreateDeviceTokens < ActiveRecord::Migration
+  def change
+    create_table :device_tokens do |t|
+      t.integer :user_id, :null => false
+      t.string :token, :null => false
+
+      t.timestamps
+    end
+    add_index :device_tokens, [:user_id, :token]
+    add_index :device_tokens, [:user_id]
```
sauloperez: This one is useless due to the left-most thingie. Postgresql can use that one above when only needing user_id

app/services/push_notifications/post_broadcaster_service.rb
```diff
@@ -0,0 +1,19 @@
+module PushNotifications
+  class UsersBroadcasterService
```
sauloperez: you're still working on it, right? because I don't see this class used.

mllocs: this class will be used once we implement the logic behind sending push notifications to users :)

db/migrate/20180221161343_create_device_tokens.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDeviceTokens < ActiveRecord::Migration
+  def change
+    create_table :device_tokens do |t|
+      t.integer :user_id, :null => false
+      t.string :token, :null => false
+
+      t.timestamps
+    end
+    add_index :device_tokens, [:user_id, :token]
```
mllocs: this should be uniq

db/migrate/20180221161343_create_device_tokens.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDeviceTokens < ActiveRecord::Migration
+  def change
+    create_table :device_tokens do |t|
+      t.integer :user_id, :null => false
+      t.string :token, :null => false
+
+      t.timestamps
+    end
+    add_index :device_tokens, [:user_id, :token], unique: true
```
sauloperez: :clap:

app/controllers/device_tokens_controller.rb
```diff
@@ -0,0 +1,19 @@
+class DeviceTokensController < PostsController
```
enricostano: Why is this controller inheriting from `PostsController`?

sauloperez: good catch

mllocs: waaaaaaat 👓

sauloperez:
If I understand this correctly we're still missing 

> A couple of services and classes to send push notifications.

right? In that case I would remove the route from this PR, merge and iterate on it an upcoming PR. It feels like all we can do as a first step for push notifications is done here.

mllocs:
Yup, I would also merge this, even if we there's nothing that sends notifications. I wouldn't remove the route though, since we could start registering mobile devices.

PR #332, opened 2018-03-14
fix test db config

Morantron:
Changes needed for https://github.com/coopdevs/timeoverflow-provisioning/pull/40 to work

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -23,7 +23,8 @@ development:
 
 test:
   <<: *defaults
-  database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_test') %>
+  host: localhost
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -23,7 +23,8 @@ development:
 
 test:
   <<: *defaults
-  database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_test') %>
+  host: localhost
+  database: 'timeoverflow_test'
```
sauloperez: Aren't these the values it was using so far? Or was the ENV var set?

Morantron: Yep, env var is set to `timeoverflow_development` D:

sauloperez: :trollface:

Issue #278, opened 2017-10-31
Fix login box


enricostano:
![2017-10-31-182416_1366x768_scrot](https://user-images.githubusercontent.com/748971/32238724-e837dca2-be68-11e7-8448-dc178ed3b1ec.png)

enricostano:
cc/ @mllocs

mllocs:
what about removing that "TimeOverflow" header text? You have that name right on the top bar, quite visible already.

enricostano:
ask @sseerrggii 😬

sauloperez:
Agree with @mllocs

sseerrggii:
Agree with @mllocs :+1: also check sizes in mobile

In my mobile resolution the input boxes (mail/password) are too big

mllocs:
👍 willdo

Matt-Yorkley:
It will be different for every screen, no? I think the css positioning for the devise modal needs to be rethought, instead of just cutting off the header text.

![screenshot from 2018-01-08 00-19-12](https://user-images.githubusercontent.com/9029026/34655933-c8a00d0c-f409-11e7-80db-6112ec06cd70.png)

sauloperez:
It's just that the modal should always fall below the nav bar, and without title, right?

Matt-Yorkley:
Agreed. Currently it seems to move up or down depending on how large it is.

sauloperez:
Now that we merged https://github.com/coopdevs/timeoverflow/pull/282 this seems like a good follow-up. Still interested @Matt-Yorkley ?

juanjochi:
I do not know if it will work but I changed the container size of 500px and put it to 649px and it seems that the whole box looks good, I put the screens.
![captura de pantalla_2018-03-09_14-23-26](https://user-images.githubusercontent.com/37113489/37210696-ee1963c4-23a9-11e8-8593-2130c636cf7e.png)
![captura de pantalla_2018-03-09_14-23-57](https://user-images.githubusercontent.com/37113489/37210709-fba2c03a-23a9-11e8-9792-7a9d01bf74f1.png)
![captura de pantalla_2018-03-09_14-25-34](https://user-images.githubusercontent.com/37113489/37210899-8d01bd56-23aa-11e8-9956-5503ec085fb3.png)

markets:
@enricostano @mllocs @sauloperez 👋 Maybe we can consider this as done by #375?

Matt-Yorkley:
[...]

PR #335, opened 2018-03-14
Mobile styles using bootstrap screen size variables

mllocs:
Just a PoC of how to use the bootstrap screen size variables to create the media queries.

As an example, I simplified the login box in mobile devices:

screen shot 2018-03-14 at 15 32 29

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -2,6 +2,9 @@
  *= require_self
  */
 
+@import "bootstrap-sprockets";
+@import "bootstrap";
+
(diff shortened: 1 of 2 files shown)


enricostano:
@mllocs do we need to test this on stagingto check we don't break web layouts?

mllocs:
Yup, I'll deploy it to staging 💪

markets:
👍 @mllocs 

Does this change fixes #278 too?

sauloperez:
@markets we miss you! Why don't you come say hello one day?😍

sseerrggii:
:+1:

PR #341, opened 2018-04-03
Expose Ruby version via Gemfile.

### LINKED ISSUE #340: Specify ruby version in Gemfile
This will make it easy to spot when someone is using a ruby version other than the one specified in  [.ruby-version](https://github.com/coopdevs/timeoverflow/blob/develop/.ruby-version). AFAIK, that doesn't stop devs from using another version when installing gems but it will update the Gemfile.lock with the version used. 

Those that use `rbenv`, `rvm` or `chruby` will benefit from [.ruby-version](https://github.com/coopdevs/timeoverflow/blob/develop/.ruby-version) given that these version managers read it.

markets:
Closes #340 

This reverts changes made in bf6853dd2c2aaa98146e2b07953559eaf6165fed, but sets the version by reading the Ruby version already defined in the `.ruby-version` file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby File.read('.ruby-version').strip
+
 gem 'rails', '~> 4.2'
 gem 'rails-i18n'
 gem "rdiscount"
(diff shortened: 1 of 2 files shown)


sseerrggii:
:+1: General test ok

PR #347, opened 2018-05-01
Introducing Persisters

mllocs:
First step on creating resources as side effects. In the future, when creating/updating a Post/Member/Transfer, it will create Events as side effects. Events will be used to create a kind of "Activity thread" section in a Time Bank. 

About the naming: https://english.stackexchange.com/questions/206893/persister-or-persistor

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -61,10 +61,18 @@ class User < ActiveRecord::Base
   end
 
   def add_to_organization(organization)
-    organization && members.
-      find_or_create_by(organization: organization) do |member|
-      member.entry_date = DateTime.now.utc
-    end
+    return unless organization
+
+    member = members.where(organization: organization).first_or_initialize
+
+    return member if member.persisted?
+
+    member.entry_date = DateTime.now.utc
+
+    persister = ::Persister::MemberPersister.new(member)
+    persister.save
+
+    return member if member.persisted?
   end
 
   def active?(organization)
--- a/spec/services/persister/post_persister_spec.rb
+++ b/spec/services/persister/post_persister_spec.rb
@@ -1,0 +1,28 @@
+require 'spec_helper'
+
+describe Persister::PostPersister do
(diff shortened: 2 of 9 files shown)


mllocs:
👧 👧

spec/services/persister/post_persister_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe Persister::PostPersister do
+  let(:organization) { Fabricate(:organization) }
+  let(:user) { Fabricate(:user) }
+  let(:category) { Fabricate(:category) }
+  let(:post) { Fabricate(:post, organization: organization) }
+
+  describe '#save' do
+    it 'saves the post' do
+      post = Offer.new(organization: organization, user: user, category: category, title: 'Title')
+      persister = ::Persister::PostPersister.new(post)
```
enricostano: Move both to `let` and reuse in examples

spec/services/persister/post_persister_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      persister.save
+
+      expect(post).to be_persisted
+    end
+  end
+
+  describe '#update_attributes' do
+    it 'updates the attributes' do
+      persister = ::Persister::PostPersister.new(post)
```
enricostano: Same

app/models/user.rb
```diff
@@ -61,10 +61,18 @@ def user
   end
 
   def add_to_organization(organization)
-    organization && members.
-      find_or_create_by(organization: organization) do |member|
-      member.entry_date = DateTime.now.utc
-    end
+    return unless organization
+
+    member = members.where(organization: organization).first_or_initialize
```
enricostano: `members` or `Member`?

mllocs: `members` would be the same as doing `Member.where(user: self, ...`

PR #348, opened 2018-05-01
Dashboard Events

mllocs:
### WAT
Using the Persisters to create Events. See https://github.com/coopdevs/timeoverflow/pull/347.

### Test
 - Create and update a post (offer and inquiries)
 - Create and update a transfer
 - Create and update a member (add and remove members for orgs basically)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,0 +1,20 @@
+class Event < ActiveRecord::Base
+  ACTIONS = ['create', 'update'].freeze
+
+  belongs_to :post
+  belongs_to :member
+  belongs_to :transfer
+
+  validates :action, inclusion: { :in => ACTIONS }, presence: true
+  validate :resource_presence
+
+  private
+
+  # Validates that only one resourece id is set
+  #
+  def resource_presence
+    return if post_id.present? ^ member_id.present? ^ transfer_id.present?
+
+    errors.add(:base, 'Specify only one resource id: `post_id`, `member_id` or `transfer_id`')
+  end
+end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -5,12 +5,12 @@ class Event < ActiveRecord::Base
   belongs_to :member
   belongs_to :transfer
 
-  validates :action, inclusion: { :in => ACTIONS }, presence: true
+  validates :action, inclusion: { in: ACTIONS }, presence: true
   validate :resource_presence
 
   private
 
-  # Validates that only one resourece id is set
+  # Validates that only one resource id is set
   #
   def resource_presence
     return if post_id.present? ^ member_id.present? ^ transfer_id.present?
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,11 +1,11 @@
 class Event < ActiveRecord::Base
-  ACTIONS = ['create', 'update'].freeze
+  enum action: { created: 0, updated: 1 }
 
   belongs_to :post
   belongs_to :member
   belongs_to :transfer
 
-  validates :action, inclusion: { in: ACTIONS }, presence: true
+  validates :action, presence: true
   validate :resource_presence
 
   private
--- a/db/migrate/20180501093846_create_events.rb
+++ b/db/migrate/20180501093846_create_events.rb
@@ -1,0 +1,13 @@
+class CreateEvents < ActiveRecord::Migration
+  def change
+    create_table :events do |t|
+      t.string :action, :null => false
+
+      t.integer :post_id
+      t.integer :member_id
+      t.integer :transfer_id
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20180501093846_create_events.rb
+++ b/db/migrate/20180501093846_create_events.rb
@@ -8,28 +8,32 @@
 #
 class CreateEvents < ActiveRecord::Migration
   def up
+    create_table :events do |t|
+      t.integer :action, null:false
+      t.integer :post_id
+      t.integer :member_id
+      t.integer :transfer_id
+      t.timestamps
+    end
+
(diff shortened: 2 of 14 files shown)


app/models/event.rb
```diff
@@ -0,0 +1,20 @@
...
+  belongs_to :post
+  belongs_to :member
+  belongs_to :transfer
+
+  validates :action, inclusion: { :in => ACTIONS }, presence: true
+  validate :resource_presence
+
+  private
+
+  # Validates that only one resourece id is set
```
enricostano: `s/resourece/resource`

app/models/event.rb
```diff
@@ -0,0 +1,20 @@
+class Event < ActiveRecord::Base
+  ACTIONS = ['create', 'update'].freeze
+
+  belongs_to :post
+  belongs_to :member
+  belongs_to :transfer
+
+  validates :action, inclusion: { :in => ACTIONS }, presence: true
```
sauloperez: :fire: that :hash: :rocket:

db/schema.rb
```diff
@@ -76,6 +76,9 @@
   add_index "documents", ["documentable_id", "documentable_type"], name: "index_documents_on_documentable_id_and_documentable_type", using: :btree
   add_index "documents", ["label"], name: "index_documents_on_label", using: :btree
 
+# Could not dump table "events" because of following StandardError
+#   Unknown type 'action_enum' for column 'action'
+
```
enricostano: :warning: a solution is to stop using `db/schema.rb` and start using `db/structure.sql`, let's talk about it

cc/ @sauloperez

enricostano: Another solution is to use the AR built in [enum](http://api.rubyonrails.org/v4.2/classes/ActiveRecord/Enum.html), basically storing an `integer` instead of a `string` (classic Rails approach with validation only at model level) or `ENUM` (DB check level I'm proposing)

db/migrate/20180501093846_create_events.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      CREATE TABLE events (
+        id          serial PRIMARY KEY,
+        action      action_enum NOT NULL,
+        post_id     integer REFERENCES posts,
+        member_id   integer REFERENCES members,
+        transfer_id integer REFERENCES transfers,
+        created_at  timestamp without time zone,
+        updated_at  timestamp without time zone,
+        CHECK(
```
sauloperez: I would only do this in SQL by means of an ALTER TABLE. This way we keep the rest using Rails' DSL, which might be easier to read for other devs.

mllocs:
[Events and Dashboard](https://trello.com/c/H0h1BeR4/82-events-and-dashboard)

db/schema.rb
mllocs: I think we are safe enough using AR enum :)

markets:
nice feature here team 👏 🔝 

As @mllocs and @sauloperez pointed out, I'd stick with simpler solutions like: 

a) string + model validation (inclusion)
b) AR `enum`

I think both options give us enough data integrity and seems an approach more easy to follow for more people.

rewritten:
I have been bitten badly by cases of inconsistent data representation. Even
at the cost of some more effort from the data integrity point of view, it
is always much better to see the same from sql and from ruby. I’d vote for
either Postgres enums (that appear as strings to the client) or just
strings with validation. The former can also suffer from the same issues as
ruby-to-sql enumerations in case of adding new items.

db/schema.rb
enricostano: There is another option using AR `enum` storing `integer` as I said AND adding check inclusion instead of PG `enum`. Will propose this option as it seems a good tradeoff to me

spec/services/persister/post_persister_spec.rb
```diff
@@ -4,25 +4,40 @@
...
-      persister.update_attributes(title: 'New title')
+    before { persister.update_attributes(title: 'New title') }
 
+    it 'updates the resource attributes' do
       expect(post.title).to eq('New title')
     end
+
+    # TODO: write better expectation
+    it 'creates an event' do
+      expect(Event.where(post_id: post.id).first.action).to eq('updated')
```
enricostano: I left this broken spec to remember to add the "updated" business logic in `PostPersister#update_attributes()`

db/migrate/20180501093846_create_events.rb
```diff
@@ -0,0 +1,39 @@
...
+  def up
+    execute <<-SQL
+      CREATE TABLE events (
+        id          serial PRIMARY KEY,
+        action      integer NOT NULL,
+        post_id     integer REFERENCES posts,
+        member_id   integer REFERENCES members,
+        transfer_id integer REFERENCES transfers,
+        created_at  timestamp without time zone,
+        updated_at  timestamp without time zone,
```
sauloperez: I left I comment about this already but it got collapsed. What about having the regular create_table and then an alter table for these checks that can't be done with AR? I think this would make it a bit more clear that we specify things AR does not provide and perhaps a bit more readable.

db/migrate/20180501093846_create_events.rb
```diff
@@ -0,0 +1,39 @@
+# This migration does not use Rails ActiveRecord ORM DSL since
+# it doesn't provide data integrity (foreign key) for polymorphic models.
```
sauloperez: I'm not sure I fully understand what you mean.

enricostano: Read the article :joy:

sseerrggii:
# Tested CRUD
*  Offer/inquiry :heavy_check_mark: 
* Transfer (only create and update, we can't delete) :heavy_check_mark: 
* Member :heavy_check_mark:

PR #350, opened 2018-05-08
Customizing bootstrap

mllocs:
- Changed the primary color of the app.
- Load bootstrap modules explicitly and only the ones that are needed.
- Upgraded sass gem.
- Removed "Tags" menu item for non admin users.

screen shot 2018-05-08 at 19 30 14

### DIFF
--- a/app/assets/stylesheets/_bootstrap-custom.scss
+++ b/app/assets/stylesheets/_bootstrap-custom.scss
@@ -1,0 +1,57 @@
+/*!
+ * Bootstrap v3.3.7 (http://getbootstrap.com)
+ * Copyright 2011-2016 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ */
+
+// Core variables and mixins
+@import "bootstrap-overrides";
+@import "bootstrap/variables";
+@import "bootstrap/mixins";
+
+// Reset and dependencies
+@import "bootstrap/normalize";
+@import "bootstrap/print";
+@import "bootstrap/glyphicons";
+
+// Core CSS
+@import "bootstrap/scaffolding";
+@import "bootstrap/type";
+@import "bootstrap/code";
+@import "bootstrap/grid";
+@import "bootstrap/tables";
+@import "bootstrap/forms";
+@import "bootstrap/buttons";
+
+// Components
+@import "bootstrap/component-animations";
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/_bootstrap-custom.scss
```diff
@@ -0,0 +1,57 @@
...
+@import "bootstrap/panels";
+@import "bootstrap/responsive-embed";
+// @import "bootstrap/wells";
+@import "bootstrap/close";
+
+// Components w/ JavaScript
+@import "bootstrap/modals";
+@import "bootstrap/tooltip";
+// @import "bootstrap/popovers";
+// @import "bootstrap/carousel";
```
sauloperez: We can simply remove the commented out ones, right?

mllocs:
[Polish the current design using Bootstrap](https://trello.com/c/hIT6kzrn/114-polish-the-current-design-using-bootstrap)

app/assets/stylesheets/_bootstrap-custom.scss
mllocs: I left those comments because it could help future contributors understand that we are not loading all the modules. It's kind of an explicit way of saying which Bootstrap parts are not being used.

sauloperez: IMO that could be a bit more clear if precede it with a block comment explaining just that.

mllocs: done

app/views/application/_navbar.html.erb
```diff
@@ -51,8 +51,8 @@
               <%= render 'application/menus/user_list_link' %>
               <%= render 'application/menus/offers_dashboard_link' %>
               <%= render 'application/menus/inquiries_list_link' %>
-              <%= render 'application/menus/offers_by_tag_link' %>
               <% if current_user.manages? current_organization %>
+                <%= render 'application/menus/offers_by_tag_link' %>
```
sauloperez: is this now only for admins?

enricostano: ping @mllocs 

cc/ @sseerrggii

mllocs: During the last redesign conversations we had it was proposed to remove this section (to simplify the interface). Instead, I'm making it available only for admins.

Issue #349, opened 2018-05-08
Don't show emails @example.com


sseerrggii:
Before angular removal #272 in the members list and member profile users with email ...@example.com show an empty email field

We want this feature back :pray:

saverio-kantox:
@sseerrggii #351 solves this :)

rewritten:
@sseerrggii #351  solves this :)

PR #351, opened 2018-05-08
Use membership decorator in view

rewritten:
Solves #349 

# What
I add here a decorator for membership. It has access to the view context, and exposes exactly what's needed to render the view.

# Why
User list view (table rows mostly) uses several models in parallel and it's difficult to follow, with all the conditionals. A decorator encapsulates all the logic in a testable class.

# Bonus
I have added dry-initializer gem, which is a very lightweight library to help with different kinds of instance initialization strategies. It is not really needed here, but it's a nice addition in general, as I would also like to introduce to more dry-* gems in the future.

* [x] Write the decorator
* [x] Specify all the decorator behavior
* [x] Modify the view so it relies on the decorated member instead of user + membership
* [x] Pass only the decorated members from UsersController#index

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,13 +2,10 @@ class UsersController < ApplicationController
   before_filter :authenticate_user!
 
   def index
-    default_sort
-
-    @search =
-      current_organization.members.joins(:account, :user).ransack(params[:q])
+    @search = current_organization.members.ransack(search_params)
 
     @members =
-      @search.result.page(params[:page]).per(25)
+      @search.result.eager_load(:account, :user).page(params[:page]).per(25)
 
     @member_view_models =
       @members.map { |m| MemberDecorator.new(m, self.class.helpers) }
@@ -63,9 +60,8 @@ class UsersController < ApplicationController
 
   private
 
-  def default_sort
-    params[:q] ||= {}
-    params[:q][:s] ||= 'member_uid asc'
+  def search_params
+    {s: 'member_uid asc'}.merge(params.fetch(:q, {}))
   end
 
   def scoped_users
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -13,12 +13,25 @@ class UsersController < ApplicationController
       .page(params[:page])
       .per(25)
 
-    @members = current_organization.members.
-      where(user_id: @users.map(&:id)).
+    user_ids = @users.pluck(:id)
+
+    @members = current_organization.
+      members.
+      where(user_id: user_ids).
       includes(:account, :user).
-      to_a.
-      sort_by { |m| @users.index(m.user) }.
-      map { |m| MemberDecorator.new(m, view_context) }
+      each_with_object({}) { |mem, ob| ob[mem.user_id] = mem }.
+      values_at(*user_ids).
+      map { |m| MemberDecorator.new(m, self.class.helpers) }
+
+    # TODO: mutate theparameters so they can be used directly on the
+    # members table.
+    #
+    # @members = current_organization.members.
+    #   joins(:account, :user).
+    #   merge(@search.result(distinct: false)).
+    #   page(params[:page]).
+    #   per(25).
+    #   map { |m| MemberDecorator.new(m, self.class.helpers) }
   end
 
   def show
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,36 +2,16 @@ class UsersController < ApplicationController
   before_filter :authenticate_user!
 
   def index
-    @search = User.ransack(params[:q])
-    @search.sorts = 'members_member_uid asc' if @search.sorts.empty?
+    default_sort
 
-    @users = @search
-      .result(distinct: false)
-      .joins(members: :account)
-      .eager_load(members: :account)
-      .where(members: { organization: current_organization.id })
-      .page(params[:page])
-      .per(25)
+    @search =
+      current_organization.members.joins(:account, :user).ransack(params[:q])
 
-    user_ids = @users.pluck(:id)
+    @members =
+      @search.result.page(params[:page]).per(25)
 
-    @members = current_organization.
-      members.
-      where(user_id: user_ids).
-      includes(:account, :user).
-      each_with_object({}) { |mem, ob| ob[mem.user_id] = mem }.
-      values_at(*user_ids).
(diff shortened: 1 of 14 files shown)


rewritten:
Lots of spec-related noise, I will squash all commits after CR is done, before merging.

app/decorators/member_decorator.rb
```diff
@@ -0,0 +1,64 @@
...
+  def manager?
+    !!member.manager
+  end
+
+  def row_css_class
+    'bg-danger' unless active?
+  end
+
+  def inactive_icon
+    v.tag(:span, class: %w[glyphicon glyphicon-time]) unless active?
```
markets: @rewritten I see this produces a different html compared to the original code: an anchor inside the span (called along with `link_to_self`). I think this is because this line returns a self-closing span. It looks like this:

captura de pantalla 2018-05-12 a la s 18 01 52

We can fix that by using the `glyph` helper:
    
    v.glyph('time') unless active?

I just tried it locally and it looks fine:

captura de pantalla 2018-05-12 a la s 18 18 42

app/controllers/users_controller.rb
```diff
@@ -13,11 +13,25 @@ def index
...
-    @memberships = current_organization.members.
-      where(user_id: @users.map(&:id)).
-      includes(:account).each_with_object({}) do |mem, ob|
-        ob[mem.user_id] = mem
-      end
+    user_ids = @users.pluck(:id)
+
+    @members = current_organization.
+      members.
+      where(user_id: user_ids).
```
sauloperez: can we rather do an AR merge? This avoids one SQL query which saves a network round trip, etc. PostgreSQL knows how to match against those users better than us.

app/decorators/member_decorator.rb
```diff
@@ -0,0 +1,64 @@
...
+    h.member_path(member)
+  end
+
+  def toggle_active_member_path
+    h.toggle_active_member_path(member)
+  end
+
+  private
+
+  def h
```
sauloperez: what about a more meaningful name like `routes`? It would help other devs understand all this.

spec/controllers/users_controller_spec.rb
```diff
@@ -67,13 +67,15 @@
...
         get :index
 
-        expect(assigns(:memberships)).to eq({
-          member.user_id => member,
-          another_member.user_id => another_member,
-          member_admin.user_id => member_admin,
-          wrong_email_member.user_id => wrong_email_member,
-          empty_email_member.user_id => empty_email_member
-        })
+        assigns(:members).each { |m| expect(m).to be_a(MemberDecorator) }
```
sauloperez: IMO it's a not a good idea to couple to particular class types. Makes future changes harder and it's sort of against Ruby's duck-type principle.

app/controllers/users_controller.rb
```diff
@@ -13,11 +13,25 @@ def index
...
+
+    # TODO: mutate theparameters so they can be used directly on the
+    # members table.
+    #
+    # @members = current_organization.members.
+    #   joins(:account, :user).
+    #   merge(@search.result(distinct: false)).
+    #   page(params[:page]).
+    #   per(25).
+    #   map { |m| MemberDecorator.new(m, self.class.helpers) }
```
sauloperez: :fire:

app/decorators/member_decorator.rb
rewritten: Or `content_tag` with empty content

spec/controllers/users_controller_spec.rb
rewritten: Ok but still somewhat I want to specify that what's sent to the view is a list of decorated objects. What about a general method `decorated_object` on the decorator? Instead of `member`. I have used `member` just for laziness, for instance when using draper I tend to use `object` instead of the named accessor.

app/controllers/users_controller.rb
rewritten: 👍

app/controllers/users_controller.rb
rewritten: PostgreSQL knows, ActiveRecord sometimes messes up the joins. I will try to see if this is working in this case.

app/controllers/users_controller.rb
rewritten: Actually I left this here to remind about making the member table ransack against members itself. I will try to complete the feature, and make the whole page based on members so we don't need the additional `@users` variable here.

app/controllers/users_controller.rb
rewritten: Indeed it is not working in this case. As `@users` already comes from a JOIN, AR is not really able to distinguish how to prepare the association chain.

Here it is a direct ransack on members, so we don't have to mix stuff up.

rewritten:
* Dry removed 😢 
* Base `ViewModel` class added, it should be more contributor-friendly
* Refactored users index action to be actually based on members (the only remnant is that it's UsersController, because it's also accessed with the `members` path...)
* Slightly tweaked index view so ransack works based on Member relation

Ready for your second review :)

app/controllers/users_controller.rb
```diff
@@ -2,22 +2,16 @@ class UsersController < ApplicationController
...
-
-    @memberships = current_organization.members.
-      where(user_id: @users.map(&:id)).
-      includes(:account).each_with_object({}) do |mem, ob|
-        ob[mem.user_id] = mem
-      end
+    default_sort
+
+    @search =
+      current_organization.members.joins(:account, :user).ransack(params[:q])
```
rewritten: There doesn't seem to be any difference between doing `joins` before `ransack.result` or after it, and I don't have a clear reason for it to be in any of them instead of the other. The only "intuition" I have is that by joining after the search, we may risk joining twice. Any input?

Maybe I should just ditch the joins and use only `eager_load`.

app/controllers/users_controller.rb
```diff
@@ -2,22 +2,16 @@ class UsersController < ApplicationController
...
+
+    @members =
+      @search.result.page(params[:page]).per(25)
+
+    @member_view_models =
```
rewritten: There is no easy way to keep the pagination info exposed and at the same time to decorate the objects. It's this or another "paginated collection decorator", which most of the times I find ugly.

rewritten:
@markets, @sauloperez what's the process now? Who merges? Who deploys?

sauloperez:
we need @sseerrggii and/or his testing crew to give it a try.

sseerrggii:
I have tested and everything works fine for me :+1:

Issue #327, opened 2018-02-15
Is this optimization really needed on password reset?


enricostano:
I still don't understand what problem we were trying to solve.

Give a look at https://github.com/coopdevs/timeoverflow/commit/f8f940050ac5677805beca4065f9cdf0f7f287bb

sauloperez:
See my comment https://github.com/coopdevs/timeoverflow/commit/f8f940050ac5677805beca4065f9cdf0f7f287bb

juanjochi:
From what I understand it seems that you have to check that when asking for password recovery, you receive an email and you can change it correctly. I have requested the recovery three times, with Opera what I have detected is that the first took several minutes, the other two times the mail arrived in seconds. On the other hand I think that to facilitate the use could hide the button to send the mail again and the unlock and appear after pressing once the password reset button. I have tried from Firefox and it takes seconds to get the recovery mail.

enricostano:
Maybe related https://github.com/coopdevs/timeoverflow/blob/develop/config/initializers/devise.rb#L187

markets:
What are you suggesting @enricostano to increase or decrease that number? Would be amazing to fix this (and remove `memcached` :stuck_out_tongue:) by just tweaking Devise settings.

Another possible approach: to sort out this by db, with something like:

```ruby
def set_reset_password_token
    return reset_password_token if reset_password_sent_at < RESET_PASSWORD_TOKEN_LAZINESS.ago
  end
```

Instead of the current:

https://github.com/coopdevs/timeoverflow/blob/e6ef11997badfcf07a714bb2a44b70757741ddf7/app/models/concerns/lazy_recoverable.rb#L11-L17

enricostano:
I would like to try out Devise settings tuning... and remove the cache.

markets:
@sseerrggii can we just delete this? I'd take care of the ✂️ but I want to confirm first ...

This was added in the past because people are getting different emails (multiple clicks) and only the last one is valid (to reset the pass). But nowadays most of emails clients have good UI for threads/groups, so maybe we can just 🔥 this optimization (=> and remove the entire cache gem and infra piece). 

We can also tune this setting if you want:

```
# Time interval you can reset your password with a reset password key.
# Don't put a too small interval or your users won't have the time to
# change their passwords.
config.reset_password_within = 6.hours
```

[...]

PR #353, opened 2018-05-10
Add background jobs

enricostano:
### WAT
We need a background jobs system in order to send emails, mobile push notifications, etc.

### Dependencies
https://github.com/coopdevs/timeoverflow-provisioning/pull/47 (Install Redis)

### TODO BEFORE DEPLOY
 - [x] provision `redis` on `staging` (Ansible scripts)
 - [x] install `redis` by hand in `production` :grimacing:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem 'prawn-table', '~> 0.2.2'
 gem 'elasticsearch-model'
 gem 'elasticsearch-rails'
 gem 'skylight'
+gem 'sidekiq', '5.1.3'
 
 # Assets
 gem 'jquery-rails', '>= 4.2.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -112,6 +112,7 @@ GEM
     coffee-script-source (1.8.0)
     columnize (0.9.0)
     concurrent-ruby (1.0.5)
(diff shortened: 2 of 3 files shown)


markets:
Nice @enricostano! Taking into account that we'll have (and we'll depend on) `redis`, we can maybe drop the `memcached` dependency and use `redis` also as a cache store, in the future, so this will simplify our infrastructure.

sauloperez:
that's a good idea @markets . I wonder though if we can completely get rid of the cache store to simplify the infrastructure as well as the code even further.

markets:
Yes, right now we're using caching (=> and `memcached`) only for this #327 I think :(

sauloperez:
Yep, that's we got it in our radar. :hocho: if possible.

enricostano:
> we can maybe drop the memcached dependency and use redis also as a cache store, in the future, so this will simplify our infrastructure.

I whish we could, I remember a conversation with sys admins about redis not being the best option for cache... but I need to dig into that again. It was probably related to performance, I'll look into it.

sauloperez:
Could be. But keep in mind that our use case is rather small. It might be one of those cases where the tradeoff of using it as a cache store makes sense.

PR #355, opened 2018-05-12
Organizations: minor code cleanup

markets:
This is just a minor code cleanup 🔨 (about Organizations):

- controller: code rearrange and simplification
- model: 
  - remove unused code
  - added `validates :name, presence: true`
- views: 
  - use `show_error_messages!(@organization)` to print error messages in form
  - "new" button with pull-right class (for consistency with all "new" buttons)
- specs: minor cleanup, fix repeated test and add new spec for name presence 

We need to properly define and implement an `OrganizationPolicy`, but this would be a different PR (we should define an access control policy first), this branch only contains some minor refactors, which clarify/simplify code and remove 💀 methods. What do you think guys? The only behavior change is the validation on name presence, which is probably a sane change.


app/controllers/organizations_controller.rb
```diff
@@ -1,20 +1,12 @@
...
-  before_filter :load_resource
-
-  def load_resource
-    if params[:id]
-      @organization = Organization.find(params[:id])
-    else
-      @organizations = Organization.all
-    end
-  end
+  before_filter :load_resource, except: [:new, :index, :create]
```
sauloperez: what about using `only` instead? It will feel a bit more under control. Loading it by default in any new action seems dangerous to me.

app/controllers/organizations_controller.rb
```diff
@@ -1,20 +1,12 @@
...
 
   def new
     @organization = Organization.new
   end
 
   def index
-    @organizations = @organizations.matching(params[:q]) if params[:q].present?
```
sauloperez: why are we removing the `q` param? That will break the search, right?

spec/models/organization_spec.rb
```diff
@@ -17,35 +17,43 @@
...
-    organization.web = ""
-    expect(organization).to be_valid
-    expect(organization.web).to eq ""
+  describe 'ensure_url validation' do
+    it "without http & https" do
+      organization.web = "www.casa.com"
+      expect(organization).to be_valid
+      expect(organization.web).to eq "http://www.casa.com"
+    end
+    it "with http" do
```
sauloperez: can we add some spacing between text examples, now that we are at it?

app/models/organization.rb
```diff
@@ -15,11 +11,7 @@ def with_transfer
   has_many :inquiries
   has_many :documents, as: :documentable
 
-  scope :matching, ->(str) {
-    where(Organization.arel_table[:name].matches("%#{str}%"))
-  }
-
-  validates :name, uniqueness: true
+  validates :name, presence: true, uniqueness: true
```
sauloperez: let's make sure we also have `null: false` and a unique index at db level 🙏

app/controllers/organizations_controller.rb
markets: I deleted it becasue I can't found any point of the app using this search.

app/models/organization.rb
markets: agree @sauloperez I'll review the schema and if in case this is missing, I'll create a migration to fix that

markets:
@sauloperez code review changes pushed!

sseerrggii:
Hi @markets I tested to create a organization, without name now it isn't possible :clap: . Is it enough test?

May be you can help with other improvements in active admin, if you want I can create the issues: 

* I can't delete a organization: If some admin ask to me to remove all timebank I delete all the members (and users if they haven't  another membership) and after that I change the name of timebank with something like "to delete someday" :pray: 

* Another annoying thing is when I create a new user and add membership the select of all organitzations show the organitzations without an order, for me will be perfect if they are ordered by ID 

:kissing_heart: :kissing_heart: :kissing_heart:

enricostano:
Thanks @sseerrggii @markets :tada: 

> May be you can help with other improvements in active admin

Remember that we have a prioritized list of stuff to do in Trello, I would rather start from there :pray:

app/models/organization.rb
enricostano: @markets can you please check this, thanks!!! (no rush of course :smile: )

enricostano: Oh, I see you did it already here https://github.com/coopdevs/timeoverflow/pull/355/commits/b9d9b9636c806d48aad7e75cd353160572750689 right?

enricostano:
@sseerrggii @markets can we give this PR as tested and merge it?

enricostano:
@markets sorry, would you mind to rebase it from the latest `develop`? Thanks!

markets:
@enricostano updated!

PR #356, opened 2018-05-13
Added push notifications setting

mllocs:
Added a "push notifications" global setting in the users table. 

Eventually, we'll need to create a user_settings table, moving all the settings from the users table to the new one. I thought that using this new approach (user_settings table) only for this new setting would actually create complexity instead of reducing it. I think that it would be better to implement user_settings once we can move all the other user attributes that are settings.

I fixed the markup of the form and removed a css hack that is not needed when the markup is correct. I also removed some code and translations that were not being used.

screen shot 2018-05-13 at 11 12 04

### DIFF
--- a/config/locales/eu.yml
+++ b/config/locales/eu.yml
@@ -500,8 +500,7 @@ eu:
     edit:
       edit_user: Erabiltzailea aldatu
     form:
-      admin_warning: Kontuz!!! Erabiltzaile honi boterea ematen ari zara.
-      superadmin_warning: Kontuz!!! Erabiltzaile honi izugarrizko boterea ematen ari zara.
+      notifications: Jakinarazpen
     give_time:
       give_time: honi denbora eman
     index:
--- a/config/locales/pt-BR.yml
+++ b/config/locales/pt-BR.yml
@@ -494,8 +494,7 @@ pt-BR:
     edit:
       edit_user: Trocar usuário
     form:
-      admin_warning: Atenção!!! Você está dando poderes a este usuário!!
-      superadmin_warning: Atenção!!! Você está dando PODERES DIVINOS a este usuário!!
+      notifications: Notificações
     give_time:
       give_time: Dar Tempo a
     index:
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -361,11 +361,6 @@ label[required]::after{
   }
 }
 
-form .checkbox input[type="checkbox"] {
-  margin-left: 5px;
-  position: relative;
-}
-
 .back-overlay {
   background-color: $black;
   background-size: cover;
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -70,7 +70,7 @@ class UsersController < ApplicationController
 
   def user_params
     fields_to_permit = %w"gender username email date_of_birth phone
-                          alt_phone active description notifications"
+                          alt_phone active description notifications push_notifications"
     fields_to_permit += %w"admin registration_number
                            registration_date" if admin?
     fields_to_permit += %w"organization_id superadmin" if superadmin?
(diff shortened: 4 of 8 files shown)


mllocs:
[Notification settings](https://trello.com/c/35dtB2cb/105-notification-settings)

config/locales/pt-BR.yml
```diff
@@ -494,8 +494,7 @@ pt-BR:
     edit:
       edit_user: Trocar usuário
     form:
-      admin_warning: Atenção!!! Você está dando poderes a este usuário!!
-      superadmin_warning: Atenção!!! Você está dando PODERES DIVINOS a este usuário!!
+      notifications: Notificações
```
sauloperez: very professional

db/migrate/20180512180546_add_push_notifications_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddPushNotificationsToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :push_notifications, :boolean, default: true
```
sauloperez: I would add `null: false`

db/schema.rb
```diff
@@ -76,6 +76,15 @@
...
 
+  create_table "events", force: :cascade do |t|
+    t.string   "action",      null: false
+    t.integer  "post_id"
+    t.integer  "member_id"
+    t.integer  "transfer_id"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
```
enricostano: Ooops :fire:

db/schema.rb
```diff
@@ -159,6 +168,13 @@
     t.integer "post_id"
   end
 
+  create_table "user_settings", force: :cascade do |t|
+    t.integer "user_id"
+    t.boolean "push_notifications_all"
+  end
+
+  add_index "user_settings", ["user_id"], name: "index_user_settings_on_user_id", using: :btree
+
```
enricostano: Oops :fire:

db/migrate/20180512180546_add_push_notifications_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddPushNotificationsToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :push_notifications, :boolean, default: true, null: false
+  end
+end
```
enricostano: I don't see any change in the `User` model.

mllocs: you gotta love Rails 😏

mllocs:
Ready @sauloperez @enricostano

db/migrate/20180512180546_add_push_notifications_to_users.rb
sauloperez: I guess he meant we still need to validate the presence of `push_notifications` to get a human-readable error message?

db/schema.rb
```diff
@@ -208,6 +208,7 @@
     t.datetime "locked_at"
     t.string   "locale",                 default: "es"
     t.boolean  "notifications",          default: true
+    t.boolean  "push_notifications",     default: true, null: false
```
enricostano: Do we need to add a data migration for the existing records?

config/locales/eu.yml
```diff
@@ -500,8 +500,7 @@ eu:
     edit:
       edit_user: Erabiltzailea aldatu
     form:
-      admin_warning: Kontuz!!! Erabiltzaile honi boterea ematen ari zara.
-      superadmin_warning: Kontuz!!! Erabiltzaile honi izugarrizko boterea ematen ari zara.
+      notifications: Jakinarazpen
```
enricostano: Should we use locale app for this? cc/ @sseerrggii

db/schema.rb
sauloperez: I believe the default makes it unnecessary.

mllocs: existing records will default to `true`

enricostano:
Rebase please @mllocs

PR #333, opened 2018-03-14
Feature specs

Morantron:
Resurrecting https://github.com/coopdevs/timeoverflow/pull/254

PR #361, opened 2018-05-19
Admin organizations order

### LINKED ISSUE #360: Active admin: user to membership organitzations order
One annoying thing in Active Admin is when superadmin creates a new user and add membership the select of all organitzations list show the organitzations list without an order

 Will be perfect if orgs are ordered by ID

markets:
Closes #360 

> One annoying thing in Active Admin is when superadmin creates a new user and add membership the select of all organitzations list show the organitzations list without an order
> 
> Will be perfect if orgs are ordered by ID

**EXTRA**

Helpers cleanup: 
- remove empty helpers to avoid loading unnecessary constants (and it's pretty easy to add it again if necessary)
- minor refactor on `avatar_url` helper
- moved alert_class to ApplicationHelper (+ minor refactor)
- use valid css error class when using `show_error_messages!(resource)` helper (possible regression introduce in https://github.com/coopdevs/timeoverflow/commit/d03726cb54654417c188a801051733385a006996)
- added some tests for ApplicationHelper and GlyphHelper


app/helpers/application_helper.rb
```diff
@@ -6,11 +6,14 @@ module ApplicationHelper
...
-    "https://www.gravatar.com/avatar/#{gravatar_id}.png?" +
-      "#{Rack::Utils.build_query(gravatar_options)}&d=identicon"
+    gravatar_options = {
+      set: "set1",
+      gravatar: "hashed",
+      size: "#{size}x#{size}",
+      d: "identicon"
+    }
+
+    "https://www.gravatar.com/avatar/#{gravatar_id}.png?#{gravatar_options.to_param}"
```
enricostano: 😍

markets:
Hi @rewritten, if I'm not wrong to use `RSpec.describe` is recommended by the RSpec team (https://relishapp.com/rspec/rspec-core/v/3-0/docs/configuration/global-namespace-dsl) and the official docs, so I'm +1 to move to this form, but I'd leave it for a different PR, so the team can discuss about pros/cons, thoughts, settings, ....

rewritten:
Works for me

spec/spec_helper.rb
```diff
@@ -67,15 +68,12 @@
   # the seed, which is printed after each run.
   #     --seed 1234
 
-  puts "Randomized with seed #{config.seed}."
-
```
enricostano: Is this printed out by default now? I find it useful.

enricostano:
@markets @rewritten I created an issue to keep track of the RSpec scope thing https://github.com/coopdevs/timeoverflow/issues/362

app/helpers/application_helper.rb
sauloperez: I don't mean to do it now but these are the kind of things that I like to turno into a value object. And this will eventually reduce the amount of helpers. Just a thought worth sharing.

spec/spec_helper.rb
markets: Yes @enricostano, I saw it duplicated now.

enricostano: Weirdly enough I don't see it duplicated in `develop` :thinking:

markets: Maybe you have a global `.rspec` with any kind of setting preventing this to be displayed?

enricostano: Not that I am aware of :joy:
```
timeoverflow@timeoverflow:/var/www/timeoverflow$ ls ~/.rspec
ls: cannot access '/home/timeoverflow/.rspec': No such file or directory
timeoverflow@timeoverflow:/var/www/timeoverflow$ ls ./.rspec
ls: cannot access './.rspec': No such file or directory
```

markets: Ok @enricostano, it's my global `.rspec` 😂

```
[2.3.0] markets: ~/workspace/timeoverflow (develop)$ more ~/.rspec 
--color
--order random
--format documentation
```

This is printed by default by `RSpec` if you have `config.order = "random"`, but this was removed here: https://github.com/coopdevs/timeoverflow/pull/333/commits/94c50fbec07ae4fcecfd7d31914456d42d5b1dc1. I'm going to put it back :sweat_smile:

app/helpers/application_helper.rb
```diff
@@ -80,4 +83,15 @@ def get_body_css_class(controller)
...
+  def alert_class(alert)
+    case alert
+    when 'error', 'alert'
+      'alert-danger'
+    when 'notice'
+      'alert-success'
+    else
+      'alert-info'
+    end
+  end
```
markets: @mllocs what do you think? specially this commit: https://github.com/coopdevs/timeoverflow/pull/361/commits/761c7ff9d38c74fae3a31a285387b36d3892e6c1 😄

enricostano:
Deployed on https://staging.timeoverflow.org/

enricostano:
@sseerrggii tested it :green_apple:

PR #364, opened 2018-05-22
Added organization_id to events

mllocs:
Little big detail, Events are org scoped, heh.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -4,6 +4,7 @@ class Event < ActiveRecord::Base
   belongs_to :post
   belongs_to :member
   belongs_to :transfer
+  belongs_to :organization
 
   validates :action, presence: true
   validate :resource_presence
--- a/app/services/persister/transfer_persister.rb
+++ b/app/services/persister/transfer_persister.rb
@@ -1,9 +1,10 @@
 module Persister
   class TransferPersister
-    attr_accessor :transfer
+    attr_accessor :transfer, :organization
 
-    def initialize(transfer)
+    def initialize(transfer, organization)
       @transfer = transfer
+      @organization = organization
     end
 
     def save
@@ -29,11 +30,11 @@ module Persister
     private
 
     def create_save_event!
-      ::Event.create! action: :created, transfer: transfer
+      ::Event.create! action: :created, transfer: transfer, organization: organization
     end
 
     def create_update_event!
-      ::Event.create! action: :updated, transfer: transfer
+      ::Event.create! action: :updated, transfer: transfer, organization: organization
     end
   end
 end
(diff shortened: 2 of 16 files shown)


app/views/events/index.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<ul>
+  <% @events.each do |e| %>
+    <li>
+      <span><%= e.id %></span>
+      <span><%= e.action %></span>
+      <span><%= e.updated_at %></span>
+      <span><%= e.created_at %></span>
+    </li>
+  <% end %>
+</ul>
```
mllocs: This is hidden for now

app/models/event.rb
```diff
@@ -4,6 +4,7 @@ class Event < ActiveRecord::Base
   belongs_to :post
   belongs_to :member
   belongs_to :transfer
+  belongs_to :organization
```
enricostano: Is this needed to `Event.where(organization_id: 27)`?

app/services/persister/transfer_persister.rb
```diff
@@ -1,9 +1,10 @@
 module Persister
   class TransferPersister
-    attr_accessor :transfer
+    attr_accessor :transfer, :organization
 
-    def initialize(transfer)
+    def initialize(transfer, organization)
       @transfer = transfer
+      @organization = organization
```
enricostano: Can't we infer from `@transfer.organization`?

app/models/event.rb
mllocs: nope, it's needed for `event.organization` and the setters (event.organization = organization), at least.

app/services/persister/transfer_persister.rb
mllocs: transfer doesn't have org id, It should be something like transfer.post.organization but then I realised that it was easier to pass it as an option to the Persister.

app/models/event.rb
enricostano: Yeah... I meant not literally the field, but 
the duplicated info in the column. if you don't need that kind of `where` you can infer the org from the associated resource

app/services/persister/transfer_persister.rb
enricostano: The less we expose the better, I vote for using `transfer.post.organization`

app/models/event.rb
sauloperez: Yep, I totally share the same view. Here we're duplicating information that will surely bite us in the shape of data inconsistency.

mllocs: well, in theory ™️ events are read only, an not likely to be moved between orgs.

PR #367, opened 2018-05-25
Push notifications background job for Post

enricostano:
### WAT
Whit this PR we add a new step in the `PostPersister` service while creating or updating a `Post`: we also enqueue now a `CreatePushNotificationsJob`.

The background job will retrieve the `Event` and the `Post` associated to the event and will create as many `PushNotification` resources as needed.

In the case of `Post` it will create a `PushNotification` for each user in the `Post`'s organization that have a `DeviceToken`. In future PRs we'll add this feature for `Member` and `Transfer` resources.

This PR only creates the `PushNotification` resources, it doesn't send any push notification to the mobile devices yet. The plan is to add a scheduled background job that will send them in batches.

### What to test
The following workflow:
 - a user creates a Post
 - in the DB will be created as many `PushNotification` records as the people in the organization with a `DeviceToken`

### TODO
 - [x] Setup Redis in Travis CI or just use `test` engine for ActiveJob


app/jobs/create_push_notifications_job.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePushNotificationsJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(event_id:)
+    event = ::Event.find_by_id(event_id)
+
+    raise 'A valid Event must be provided' unless event
+
+    ::PushNotifications::Creator.create!(event: event)
```
sauloperez: Do we really need the leading `::`? This reference lives in the global namespace so I don't see why `PushNotifications::Creator` would have any problem finding the constant.

app/models/push_notification.rb
```diff
@@ -0,0 +1,5 @@
+class PushNotification < ActiveRecord::Base
+  belongs_to :event, foreign_key: 'event_id'
+
+  validates_presence_of :event
```
sauloperez: Please, use the (latest) [validates](http://guides.rubyonrails.org/active_record_validations.html#presence) syntax instead. We don't know how long the old syntax will be maintained.

spec/models/push_notification_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+RSpec.describe PushNotification do
+  describe 'Validations' do
+    it { is_expected.to validate_presence_of(:event) }
+  end
+
+  describe 'Relations' do
```
sauloperez: I'd rather name it `association` which is the convention from Rails. The way I see it we're are testing it from the AR's perspective.

app/services/push_notifications/event_notifier/post_logic.rb
```diff
@@ -0,0 +1,24 @@
+module PushNotifications
+  class EventNotifier
```
sauloperez: I don't follow why we're nesting this `PostLogic` here? what's the actual reasoning? What confuses me I think is the fact that `EventNotifier` is both a class and a namespace 🤔

app/services/push_notifications/event_notifier/post_logic.rb
```diff
@@ -0,0 +1,24 @@
...
+    class PostLogic
```
sauloperez: I would try to find a better name for this. `PostLogic` is rather vague and as a reader, doesn't tell me much of what to expect from this class.

app/services/push_notifications/event_notifier/post_logic.rb
```diff
@@ -0,0 +1,24 @@
...
+      def initialize(event:)
+        @event = event
+        @post = event.post
+      end
+
+      # Conditions for Post:
+      #  - All the members of the associated organization
+      #  - that have a DeviceToken associated
```
sauloperez: this second bullet point looks like the continuation of the previous one ⚠️ . It isn't needed, right?

spec/services/push_notifications/event_notifier/post_logic_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:event) { Fabricate.build(:event, post: post, action: :created) }
+
+  describe '#users' do
+    context 'when a user pertains to the Post\'s organization' do
+
+      before { organization.members.create(user: user) }
+
+      context 'and the user has a DeviceToken associated' do
+
+        before { Fabricate(:device_token, user: user, token: 'aloha') }
```
sauloperez: the spacing up until here makes it a bit hard to read. I'd simply keep the before's right after their context.

spec/services/push_notifications/event_notifier/post_logic_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        it 'returns the user' do
+          expect(described_class.new(event: event).users).to include(user)
+        end
+      end
+
+      context 'but the user has not a DeviceToken associated' do
```
sauloperez: ⚠️ has no Device Token

app/services/push_notifications/event_notifier.rb
```diff
@@ -0,0 +1,24 @@
...
+    private
+
+    attr_accessor :event
+
+    def resource_logic
+      return PostLogic.new(event: event) if event.post_id
+
+      raise 'The resource associated to the Event is not supported'
+    end
+  end
```
sauloperez: To me, this feels like a Factory. It knows what class to instantiate and but it also delegates `#users` to the underlying class. 

We can avoid this second unnecessary responsibility from `EventNotifier` by limiting it to return the `PostLogic` instance. This class is the one that actually implements `#users`. So something like:


```ruby
class Creator
  def create!
    event_notifier = EventNotifierFactory.notification_for(event: event)
    event_notifier.users.each do
      PushNotification.create!(event: event)
    end
  end
end
```

The idea here is that all kinds of `event_notifier` we come up with conform to the duck-type by implementing `#users`.

app/jobs/create_push_notifications_job.rb
enricostano: I'm just used to it, it makes it easier to ruby to find constants. Especially in Rails the constants lookup can be really "kafkian"...

app/services/push_notifications/event_notifier/post_logic.rb
enricostano: That's correct, I could have put `PostLogic` calss even in the same file where `EventNotifier` is defined... but I don't like it much. `PostLogic` basically is a local business domain, only used inside `EventNotifier`... at least for now...

app/services/push_notifications/event_notifier/post_logic.rb
enricostano: Yeah, I don't like the name either... but I wanted to move forward... let's think about a better name later.

app/services/push_notifications/event_notifier/post_logic.rb
enricostano: I'll rewrite this doc, it's an `AND` basically...

app/services/push_notifications/event_notifier.rb
enricostano: Yeah, we need to play a bit more with the relationship between `PostLogic` and `EventNotifier`

enricostano: Look at `PostLogic` class as a way to organize things inside `EventNotifier` business logic. Users outside `EventNotifier` shouldn't know anything about it, they just ask `#users` to `EventNotifier`.

sauloperez:
Looks very good :ok_hand:

enricostano:
Tested on staging.

PR #370, opened 2018-05-28
fix statistics_inactive_users html

markets:
Little 🐛 in `statistics_inactive_users` view (Latest Firefox).

Fix: render `tr` inside `tbody`.

Before (note the weird effect: borders appear on hover and not on initial render):
stats1

After:
stats2


enricostano:
Tested on `staging` :+1:

PR #374, opened 2018-05-31
Add sidekiq-cron

enricostano:
We need to execute cron jobs and since we're using Sidekiq for background jobs we'll use https://github.com/ondrejbartas/sidekiq-cron

We also add a new queue to the `default` one, called `cron`.


enricostano:
Tested in staging :+1:

PR #377, opened 2018-06-04
Scheduled job to send push notifications for Post resource

enricostano:
### WAT
The workflow is the following:

1. Every time a `Post` is created or updated we create an `Event`
2. For each `Event` we create as many `PushNotification` as needed
3. Every 5 minutes the `SendPushNotificationsJob` collect all the `PushNotification` that have not been processed yet and it sends them

### Test
 - All members with a mobile device associated to their account will receive a push notification for `create` and `update` of offers a inquiries

### TODO
 - [x] Mark `PushNotification` as processed
 - [x] Add `body` to notifications https://github.com/coopdevs/timeoverflow/blob/feature/scheduled-jobs/app/services/push_notifications/broadcast.rb#L35
 - [x] Check notifications `title`
 - [x] When clicking on the notification the app should go to the Post's URL


app/services/push_notifications/broadcast.rb
```diff
@@ -0,0 +1,28 @@
+module PushNotifications
+  class Broadcast
+    def initialize(push_notifications:)
+      @push_notifications = push_notifications
+    end
+
+    def send
+      return unless push_notifications.any?
```
sauloperez: You told me `Exponent::Push::Client` does not check this and so it blows up if `notifications` is empty, right? It would make things simpler if they did.

app/services/push_notifications/broadcast.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    private
+
+    attr_reader :push_notifications
+
+    def notifications
+      push_notifications.map do |push_notification|
+        {
+          'to' => push_notification.to,
+          'title' => push_notification.title
```
sauloperez: does `Exponent::Push::Client` requires the keys to be strings? It'd be great if we could use symbols.

config/schedule.yml
```diff
@@ -0,0 +1,4 @@
+send_push_notifications_job:
+  cron: '*/5 * * * *'
```
sauloperez: can you remind me the syntax? :trollface:

app/services/push_notifications/broadcast.rb
enricostano: Yep, but they don't :(

app/services/push_notifications/broadcast.rb
enricostano: It should be bossible :+1:

config/schedule.yml
enricostano: It's every 5 minute, but we still have to take a decision about this.

Gemfile.lock
```diff
@@ -129,28 +129,32 @@ GEM
     dotenv (1.0.2)
     dotenv-rails (1.0.2)
       dotenv (= 1.0.2)
-    elasticsearch (1.0.8)
-      elasticsearch-api (= 1.0.7)
-      elasticsearch-transport (= 1.0.7)
-    elasticsearch-api (1.0.7)
+    elasticsearch (5.0.5)
```
sauloperez: we need to take extra care about search when we test this PR :see_no_evil:

app/services/push_notifications/broadcast.rb
```diff
@@ -0,0 +1,29 @@
...
+    def initialize(push_notifications:)
+      @push_notifications = push_notifications
+    end
+
+    def send
+      return unless push_notifications.any?
+
+      client = Exponent::Push::Client.new
+
+      push_notifications.update_all(processed_at: Time.now.utc)
```
sauloperez: I'd rather update the `processed_at` after publishing the notifications just in case that fails.

enricostano: We discussed about this and it's simpler this way, push notifications are not really important. You can loose one.

enricostano:
> that Push Notifications thing is almost complete no? Maybe we should integrate #381 first (now that we have #356 also merged) and we'll ready to :shipit:

@markets done here https://github.com/coopdevs/timeoverflow/pull/408

db/schema.rb
```diff
@@ -164,6 +164,7 @@
     t.datetime "processed_at"
     t.datetime "created_at",      null: false
     t.datetime "updated_at",      null: false
+    t.string   "title",           null: false
```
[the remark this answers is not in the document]
enricostano: We need to store it in the notification. The `Event` doesn't understand anything about notifications.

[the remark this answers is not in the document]
enricostano: We'll need to do the same for `body`

app/jobs/send_push_notifications_job.rb
```diff
@@ -0,0 +1,11 @@
+class SendPushNotificationsJob < ActiveJob::Base
+  queue_as :cron
+
+  def perform
+    push_notifications = PushNotification.where(processed_at: nil)
```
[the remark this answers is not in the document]
enricostano: We'll use the `processed_at` to do that.

app/models/push_notification.rb
```diff
@@ -2,5 +2,9 @@ class PushNotification < ActiveRecord::Base
   belongs_to :event, foreign_key: 'event_id'
   belongs_to :device_token, foreign_key: 'device_token_id'
 
-  validates :event, :device_token, presence: true
+  validates :event, :device_token, :title, presence: true
+
+  def to
```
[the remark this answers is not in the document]
enricostano: wat?

app/services/push_notifications/creator.rb
```diff
@@ -11,12 +11,22 @@ def initialize(event:)
     def create!
       event_notifier = EventNotifierFactory.new(event: event).build
       event_notifier.device_tokens.each do |device_token|
-        PushNotification.create!(event: event, device_token: device_token)
+        PushNotification.create!(
+          event: event,
+          device_token: device_token,
+          title: title
```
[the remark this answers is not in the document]
enricostano: No, it's not. We'll need to do the same for body also.

app/services/push_notifications/broadcast.rb
```diff
@@ -0,0 +1,56 @@
...
+        uri.request_uri,
+        notifications.to_json,
+        headers
+      )
+
+      unless response.is_a? Net::HTTPOK
+        raise PostError, "HTTP response: #{response.code}, #{response.body}"
+      end
+
+      push_notifications.update_all(processed_at: Time.now.utc)
```
mllocs: would you update `updated_at` as well?

app/jobs/send_push_notifications_job.rb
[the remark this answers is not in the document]
mllocs: there's a limit of 100 messages per request: https://docs.expo.io/versions/latest/guides/push-notifications#http2-api

mllocs:
Relevant: https://blog.expo.io/upcoming-changes-to-the-expo-push-notification-service-summer-2018-254030da786b

db/migrate/20180604145622_add_title_to_push_notification.rb
```diff
@@ -0,0 +1,10 @@
+class AddTitleToPushNotification < ActiveRecord::Migration
+  def up
+    add_column :push_notifications, :title, :string, null: false, default: 'Title for existing records'
+    change_column_default(:push_notifications, :title, nil)
```
enricostano: Why this? We already set a default in the previous line, also we have a not null constrain.

db/migrate/20180828160700_add_body_to_push_notification.rb
```diff
@@ -0,0 +1,10 @@
+class AddBodyToPushNotification < ActiveRecord::Migration
+  def up
+    add_column :push_notifications, :body, :string, null: false, default: 'Body for existing records'
+    change_column_default(:push_notifications, :body, nil)
```
enricostano: Why this? We already set a default in the previous line, also we have a not null constrain.

db/migrate/20180831161349_add_data_to_push_notification.rb
```diff
@@ -0,0 +1,10 @@
+class AddDataToPushNotification < ActiveRecord::Migration
+  def up
+    add_column :push_notifications, :data, :json, null: false, default: '{}'
+    change_column_default(:push_notifications, :data, nil)
```
enricostano: Why this? We already set a default in the previous line, also we have a not null constrain.

db/migrate/20180604145622_add_title_to_push_notification.rb
mllocs: hehehe, this code is yours and was about to ask the same 
:D https://github.com/coopdevs/timeoverflow/pull/377/commits/8069d5173007b39cc6bf37aa02b64f47846f291e#r193443513

mllocs:
The mobile app part: https://github.com/coopdevs/timeoverflow-mobile-app/pull/3

db/migrate/20180604145622_add_title_to_push_notification.rb
enricostano: THAT'S IMPOSSIBLE

PR #380, opened 2018-06-13
Fixed "Remember me" in Login

mllocs:
giphy

Here I'm trying to make the "Remember me" option work, which will be the default option in the mobile app. 

Instead of expiring the session *the hard way*, I'm using the `timeoutable` devise module. I'm also configuring the `rememberable` module to remember sessions for 4 weeks. 

I also did a little cleanup on the `alert` bootstrap component, so, instead of using our own styled alert messages, I'm using bootstrap ones, which look exactly the same :D 

I also updated devise from 4.4.1 to 4.4.3 🚀 

# UPDATE

How to test: 

Change `config.remember_for` and `config.timeout_in` in your environment. In the gif, timeout_in was set to 5.seconds and remember_for to 10.seconds. With those values:

- If I DON'T check Remember me, whenever I stay more than 5 seconds without doing a request, I'll be kicked out. 
- if I check Remember me, I can use the app without doing login again for 10 seconds. 

In the gif you can also appreciate another case:

- When you are kicked out and you go to a public page (like offers or demands), you'll see a flash message saying that your session expired but you are not redirected to the login. 

- Instead, if you navigate to a page where you need to be looged in to see, whenever your session expires you'll be redirected to the login page and you'll see the flash message there as well.


app/views/layouts/_messages.html.erb
```diff
@@ -1,5 +1,7 @@
 <% flash.each do |key, value| %>
-  <div class="alert alert-<%= key %>">
+  <% next if key == 'timedout' %>
```
mllocs: related: https://github.com/plataformatec/devise/issues/1777

app/controllers/posts_controller.rb
```diff
@@ -1,4 +1,6 @@
 class PostsController <  ApplicationController
+  before_action :authenticate_user!
```
mllocs: 👋

mllocs:
This test fails `OffersController GET #show with valid params without a logged in user assigns the requested offer to @offer`. Do we want to allow logged out visitors to see individual offers? 🤔 

cc @sseerrggii @enricostano

sseerrggii:
Yes, offers and demands (Post) are public, the view of post changes if you are logged in or logged out.

app/controllers/posts_controller.rb
enricostano: :thinking: no, no? :joy:

app/helpers/messages_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+      'alert-danger'
+    elsif alert == 'success'
+      'alert-success'
+    elsif alert == 'notice'
+      'alert-info'
+    else
+      'alert-info'
+    end
+  end
+end
```
markets: @mllocs maybe we can move this to `ApplicationHelper`, I think we don't need an extra module only for this helper method (I'm also removing "blank" helpers in #361).

app/views/layouts/_messages.html.erb
```diff
@@ -1,5 +1,7 @@
...
+
+  <div class="alert <%= alert_class(key) %>">
     <button type="button" class="close" data-dismiss="alert">x</button>
       <ul>
```
markets: Can we fix indentation here 🙏 ?

enricostano:
@mllocs I'm testing this branch in staging, I set `config.remember_for = 15.seconds` and `config.timeout_in = 5.seconds`

When `remember me`:
 - IS NOT checked: it works
 - IS checked: I'm kicked out after 5 seconds of inactivity

app/views/layouts/_messages.html.erb
mllocs: I'll do it in the next PR 💪

enricostano:
Tested, it actually works :grimacing:

app/helpers/messages_helper.rb
mllocs: I'll create an issue where we can discuss this, I prefer to tackle it in a different PR since we already tested it 😄

markets: no prob @mllocs, I'll do it in #361 while rebasing develop

PR #382, opened 2018-06-20
Fix regression switching between organizations

### LINKED ISSUE #383: Switching between organizations doesn't work anymore
I guess the regression was introduced by https://github.com/coopdevs/timeoverflow/commit/b9d9b9636c806d48aad7e75cd353160572750689#diff-33bef3083bae5b2e2bea316eab44c9fdL2

We basically excluded `set_current` action from the `before_filter` conditions.

enricostano:
## WAT
Closes #383, adding `set_current` action to `load_resource` before filter whitelist.

## How to test
1. log in with `admin2@timeoverflow.org`, the user that pertains to 2 time banks
2. click on the name of the other time bank and check that it actually switches to it

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -41,6 +41,8 @@ class OrganizationsController < ApplicationController
     redirect_to organizations_path, notice: "deleted"
   end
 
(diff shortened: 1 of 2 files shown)


enricostano:
Tested on staging :green_apple:

PR #385, opened 2018-06-21
Fix login panel vertical positioning issue

Matt-Yorkley:
Closes #278.

A few CSS tweaks so the vertical positioning of the login modal doesn't jump around when it's inner content changes in height.

PR #388, opened 2018-06-23
Optimised first pages on mobile

mllocs:
Use a smaller background images on mobile devices and optimised the current one. 

- On desktop we went from 1,9MB to 1MB
- On mobile from 1,9MB to 333KB

Other minor details to improve the look of the "first pages" (Home and Login).


mllocs:
[Precarrega, loading... #spike](https://trello.com/c/d8oQYbuD/128-precarrega-loading-spike)

markets:
good job @mllocs! 👍 do you think this branch overlaps with #385 (a fix for the classic #278.)? I think so, maybe we can try to coordinate with @Matt-Yorkley in order to port/integrate his changes (if needed) here? What do you think guys?

mllocs:
Hey @markets, yes, it totally overlaps with #385, good catch. 

Here I'm also fixing other issues, sorry for mixing it up..., here I'm getting rid of `.vertical-align` because it was only used in two places, and they didn't have much in common. @Matt-Yorkley can you take a look? 😄

app/assets/stylesheets/_footer.scss
```diff
@@ -63,6 +63,7 @@
     & {
       background-color: transparent;
       border: 1px solid white;
+      padding: 4px 8px
```
markets: :end: `;`

Matt-Yorkley:
@mllocs Ok, so my other PR for the login modal might not be needed then. I notice you've used `top: 20%` for the desktop view, and `top: 60px` for the small/mobile view. Is it worth putting it at a fixed position on both, so that it's position is always the same relative to the navigation bar? Like just having `top: 60px` for both screen sizes?

enricostano:
Deployed on staging, not sure about performance improvement but I found the following zoomed background image on mobile:

screenshot_20180703-113514

Is that normal @mllocs ?

mllocs:
@enricostano good catch, fixed ;) 

@Matt-Yorkley on your suggestion, I tried to use the same percentage in the top attribute for both big screens and mobile but it didn't look well, on big screens there was too much empty space :)

enricostano:
Tested on `staging` :green_apple:

PR #392, opened 2018-07-18
Some test with members list

sseerrggii:
Only to show what I mean with the possibility of create one new view for users and move the actual view "table style" only for admin porpuoses. 

Obviously very improvable design :innocent:  but responsive! phone numbers and mail are not showing in mobile

I think the only thing that does not work is the search engine that redirects to the old view

fireshot capture 96 - banco de tiempo local - http___10 0 3 168_3000_members_index2

cc @mllocs @enricostano

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -11,6 +11,16 @@ class UsersController < ApplicationController
       @members.map { |m| MemberDecorator.new(m, self.class.helpers) }
   end
 
+  def index2
+    @search = current_organization.members.ransack(search_params)
+
+    @members =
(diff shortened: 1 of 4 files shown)


sauloperez:
Yes, that might easier to work with than a table.

sseerrggii:
What do you think @mllocs @enricostano ?

mllocs:
@sseerrggii what about the sorting feature? With this design you can't sort by any of the sortable fields.

sseerrggii:
Sorting is only needed for admin porposes

enricostano:
Closing in favor of #396

PR #393, opened 2018-07-19
add logo ajuntament BCN and Mobile app info,

sseerrggii:
fireshot capture 97 - - http___10 0 3 168_3000_pages_about

Also fix locale key name for donate link, for consistency with the rest


app/views/pages/about.html.erb
```diff
@@ -75,10 +75,28 @@
...
+          t("pages.about.donate-link"),
           "https://www.teaming.net/poreldesarrollodelosbancosdetiempo",
         )
       ).html_safe %></p>
     </div>
   </div>
+  <div class="row">
+    <div class="col-sm-10 col-sm-offset-1 donate">
+      <h2><%= t("pages.about.app-mobile") %></h2>
+      <p><%== t(
```
sauloperez: There's no `<%==` in ERB it should be `<%=` instead (check out "Recognized Tags" in http://ruby-doc.org/stdlib-1.9.3/libdoc/erb/rdoc/ERB.html). Can you please, update the other blocks in this file?

PR #395, opened 2018-08-08
Fixed login vertical alignment

mllocs:
Should fix https://github.com/coopdevs/timeoverflow/issues/278

Improved the way we position the login box. Now it is vertically centered, so it plays better in screens that are not very high. 

http://g.recordit.co/rfv6n3arfF.gif

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -405,13 +405,14 @@ label[required]::after{
 
   .login-wrapper {
     position: absolute;
-    top: 20%;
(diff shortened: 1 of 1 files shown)


sseerrggii:
I noticed this problem

peek 16-08-2018 11-48

enricostano:
Testing in staging:

screen shot 2018-08-22 at 14 36 24

cc/ @mllocs

enricostano:
Talked offline with @mllocs , it works fine for now. Merging.

PR #396, opened 2018-08-08
Members non admin page

mllocs:
http://g.recordit.co/W8w72C946m.gif

Another approach for this: https://github.com/coopdevs/timeoverflow/pull/392


enricostano:
From what I understand admins will still see the sortable and searchable good old table, but non-admins will see the simplified div-based view. Same in laptop and mobile, right @mllocs ?

What do you think @sseerrggii ?

sseerrggii:
To me seems more useful show the new view for both admin/users so everyone can enjoy it! And add the view "old table" to a menu only visible to admin, into Manage-> Members

sseerrggii:
I think can be a good moment to introduce user description in this view. 
With user description we allow users to introduce himselfs and I think that can be good point  to see this information in the members list. Splitting text if its too long

mllocs:
See https://github.com/coopdevs/timeoverflow/pull/406

PR #399, opened 2018-08-13
Re-implement the member_uid searcher

### LINKED ISSUE #398: member_uid order
Related with https://github.com/coopdevs/timeoverflow/pull/354

Now is ordering by ID alphabetically but we need numerically order

I only see it in produccion, with more than 99 members :sweat_smile: 
fireshot capture 2 - banc del temps de gracia - https___www timeoverflow org_members

markets:
Re-implement the member_uid searcher by avoiding the column casting thing (gives some problems when sorting the "casted" column as an string). The new strategy: combine both fields in the controller by manupilating the query parameters sent to Ransack.

Closes #398 
Closes #400

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -60,7 +60,14 @@ class UsersController < ApplicationController
   private
 
   def search_params
-    {s: 'member_uid asc'}.merge(params.fetch(:q, {}))
+    q = params.fetch(:q, {})
+
+    if params[:q].present?
+      q[:member_uid_eq] = q[:user_username_or_user_email_contains]
+      q[:m] = 'or'
+    end
+
+    { s: 'member_uid asc' }.merge(q)
   end
 
   def scoped_users
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -1,9 +1,4 @@
 class Member < ActiveRecord::Base
-  # Cast the member_uid integer to a string to allow pg ILIKE search (from Ransack *_contains)
-  ransacker :member_uid do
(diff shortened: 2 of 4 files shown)


mllocs:
Hello @markets! One question, in the `schema.rb` it looks like the `member_uid` column is already an integer, shouldn't Ransack work out of the box like it does with other integer fields?

markets:
Hi @mllocs, I'll try to explain this ... :sweat_smile: 

In current `develop` if you remove the `ransacker` (which makes the cast from integer to string), and you try to use the search feature it gives you this error:

```
PG::UndefinedFunction - ERROR:  operator does not exist: integer ~~* integer
LINE 1: ...."email" ILIKE '%asd%') OR "members"."member_uid" ILIKE 0)  ...
                                                             ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

Seems like `postgresql` doesn't support `ILIKE` filters over integer columns. So, in a first try, I implemented a `ransacker` to make a cast, but then this `ransacker` is used by Ransack also in the `sort_link` helper, so now it sorts the list "alphabetically" by a string :(

In summary, this new solution consists of: copy value from original `_cont` param to a new `_eq` param (`_eq` works for integers, since it doesn't use `ILIKE` under the hood), and combine them with `OR` before sending this params to Ransack.

mllocs:
what about: https://github.com/activerecord-hackery/ransack/issues/765

markets:
@mllocs cool, I'll make a try and send new PR, so we can compare and pick better solution. Thanks!

markets:
@mllocs new proposal in #400

PR #402, opened 2018-08-15
Remove jquery-ui

markets:
:fire: `jquery-ui` we only use the datepicker widget, and only in one place (`statistics_global_activity` view), so maybe it's fine to replace it with [HTML 5 date input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) and reduce assets bundle size (`jquery-ui` it's big and old :smile:).

**NEW LOOK**

captura de pantalla 2018-08-15 a la s 19 53 01


sseerrggii:
@markets Can change translation of datapicker with locale ?

Issue #381, opened 2018-06-20
Check if the user wants to receive mobile notifications


enricostano:
Related to https://github.com/coopdevs/timeoverflow/pull/356

sauloperez:
What is this exactly? We want forcefully ask all current users?

enricostano:
We just need to filter out the people that set the `push_notifications` setting to `false`.

PR #408, opened 2018-08-20
Add conditions to PushNotification creation for Post

enricostano:
### WAT
We only want to notify users that:
 - are active members of an organization
 - have `push_notifications` flag set to `true`

Fix https://github.com/coopdevs/timeoverflow/issues/381


sseerrggii:
Tested!! :+1:

PR #421, opened 2018-09-01
don't allow transfers of 0 hours and 0 minutes

### LINKED ISSUE #345: Don't allow payments of 0 hours 0 minutes
Payments of empty time should not be allowed

markets:
Closes #345 

I also removed :scissors: a couple of unused methods and added front-end validation:

captura de pantalla 2018-09-01 a la s 19 28 46

### DIFF
--- a/app/assets/javascripts/give_time.js.coffee
+++ b/app/assets/javascripts/give_time.js.coffee
@@ -11,7 +11,7 @@ giveTimeReadyFn = () ->
       if amount > 0
         form.submit()
       else
-        $(form).find('input:submit').after(' <label class="error">Time must be greater than 0</label>')
+        $(form).find('.form-actions .error').removeClass('invisible').show()
 
   $( "#new_transfer" ).validate(config)
 
--- a/app/assets/javascripts/give_time.js.coffee
+++ b/app/assets/javascripts/give_time.js.coffee
@@ -5,8 +5,13 @@ jQuery.validator.addMethod "either-hours-minutes-informed", ((value, element) ->
 giveTimeReadyFn = () ->
   config =
     submitHandler: (form) ->
-      $(" #transfer_amount ").val($(" #transfer_hours ").val() * 3600 + $(" #transfer_minutes ").val() * 60)
-      form.submit()
+      amount = $("#transfer_hours").val() * 3600 + $("#transfer_minutes").val() * 60
+      $("#transfer_amount").val(amount)
+
+      if amount > 0
+        form.submit()
+      else
+        $(form).find('input:submit').after(' <label class="error">Time must be greater than 0</label>')
 
   $( "#new_transfer" ).validate(config)
 
--- a/app/models/transfer.rb
+++ b/app/models/transfer.rb
@@ -17,6 +17,7 @@ class Transfer < ActiveRecord::Base
   has_many :movements, dependent: :destroy
   has_many :events, dependent: :destroy
 
+  validates :amount, numericality: { greater_than: 0 }
   validate :different_source_and_destination
 
   after_create :make_movements
@@ -24,14 +25,6 @@ class Transfer < ActiveRecord::Base
   def make_movements
     movements.create(account: Account.find(source_id), amount: -amount.to_i)
     movements.create(account: Account.find(destination_id), amount: amount.to_i)
-  end
-
-  def movement_from
-    movements.detect {|m| m.amount < 0 }
-  end
-
-  def movement_to
-    movements.detect {|m| m.amount > 0 }
   end
 
   def source_id
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -494,6 +494,8 @@ en:
       minute:
         one: "%{count} minute"
         other: "%{count} minutes"
+    new:
+      error_amount: Time must be greater than 0
   users:
     edit:
       edit_user: Update user
--- a/spec/controllers/transfers_controller_spec.rb
+++ b/spec/controllers/transfers_controller_spec.rb
@@ -207,5 +207,22 @@ RSpec.describe TransfersController do
         end
       end
     end
+
+    context 'with invalid params' do
+      let(:user) { member_giver.user }
+      let(:referer) { "/transfers/new?destination_account_id=#{member_taker.account.id}" }
+
+      before do
+        request.env["HTTP_REFERER"] = referer
+      end
+
+      it 'does not create any Transfer and redirects to :back if the amount is 0' do
+        expect {
+          post(:create, transfer: { amount: 0, destination: member_taker.account.id })
+        }.not_to change(Transfer, :count)
+
+        expect(response).to redirect_to(referer)
+      end
+    end
   end
 end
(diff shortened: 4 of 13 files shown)


app/models/transfer.rb
```diff
@@ -17,6 +17,7 @@ class Transfer < ActiveRecord::Base
   belongs_to :operator, class_name: "User"
   has_many :movements
 
+  validates :amount, numericality: { greater_than: 0 }
```
sauloperez: we should also add the constraint at DB level. A Postgres [CHECK](https://www.postgresql.org/docs/9.4/static/ddl-constraints.html) constraint will do.

spec/controllers/transfers_controller_spec.rb
```diff
@@ -207,5 +207,22 @@
         end
       end
     end
+
+    context 'with invalid params' do
+      let(:user) { member_giver.user }
+      let(:referer) { "/transfers/new?destination_account_id=#{member_taker.account.id}" }
+
+      before do
+        request.env["HTTP_REFERER"] = referer
```
sauloperez: I'd restore its value afterward so as not to mess up with test suite's state and avoid future headaches.

app/models/transfer.rb
markets: This model doesn't have this column, it was defined in an `attr_accessor` some lines above. I just realized the `amount` column is defined in the `movements` table, I'll add it there then.

spec/controllers/transfers_controller_spec.rb
markets: @sauloperez do you mean something like this?

```diff
diff --git a/spec/controllers/transfers_controller_spec.rb b/spec/controllers/transfers_controller_spec.rb
index c075091..5aed2e7 100644
--- a/spec/controllers/transfers_controller_spec.rb
+++ b/spec/controllers/transfers_controller_spec.rb
@@ -213,9 +213,14 @@ RSpec.describe TransfersController do
       let(:referer) { "/transfers/new?destination_account_id=#{member_taker.account.id}" }
 
       before do
+        @prev_referer = request.env["HTTP_REFERER"]
         request.env["HTTP_REFERER"] = referer
       end
 
+      after do
+        request.env["HTTP_REFERER"] = @prev_referer
+      end
+
```

app/models/transfer.rb
markets: I'm trying this: 

```ruby
execute 'ALTER TABLE movements ADD CONSTRAINT non_zero_amount CHECK(amount != 0)'
```

and I got:

```
StandardError: An error has occurred, this and all later migrations canceled:

PG::CheckViolation: ERROR:  check constraint "non_zero_amount" is violated by some row
: ALTER TABLE movements ADD CONSTRAINT non_zero_amount CHECK(amount != 0)
```

Do you have any suggestion or idea to fix this situation? I found the `NOT VALID` statement [in the docs](https://www.postgresql.org/docs/9.4/static/sql-altertable.html):

>  If the constraint is marked NOT VALID, the potentially-lengthy initial check to verify that all rows in the table satisfy the constraint is skipped

sauloperez: According to the AR validation, the DB check should be `amount > 0`, right? the problem is that there are values already in that table that violate the constraint. So we'll need to fix those cases through a data migration before applying the check...

spec/controllers/transfers_controller_spec.rb
sauloperez: something alike, but I believe [around](https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks) is better suited for that.

markets: I tried it first with `around`, but didn't get it working (you can't share @ivars, the `request` object returns nil at that point). 

To be honest, I have never used this type of reset the referer header, in larger suites. It didn't sound to me having problems with this, isn't reset automatically on each request? am I missing something 😅?

app/models/transfer.rb
markets: Hi Pau :wave: I'm going to try to summarize current situation:

The `amount` attribute in Transfer is virtual (`attr_accesor`), it requires to be `> 0`. The `amount` attribute in Movement is persisted, it requires to be `!= 0`, must allow positives and negatives:

https://github.com/coopdevs/timeoverflow/blob/dbf1a65a4b886b78ff670fb99700f5e2ab60f617/app/models/transfer.rb#L24-L27

I initially only added the check in code (`amount > 0`), in Transfer, and I'm going to add db constraint (`amount != 0`) in Movement table. Of course, actually there are movements with `amount == 0` in that table, so an error is raised.

Possible solutions:

a) use the `NOT VALID` statement, to skip the initial check (should work, didn't tried yet)
b) data migration: what to do with those records? should we remove them? ¯\_(ツ)_/¯

spec/controllers/transfers_controller_spec.rb
sauloperez: Fuuu right it's the request object so pointless comment :see_no_evil:

sseerrggii:
@sauloperez @enricostano Could you please review this?

sseerrggii:
We can delete the records with 0 hours and 0 minutes

markets:
Pushed migration with (https://github.com/coopdevs/timeoverflow/pull/421/commits/f7e47d5890135422a0ce9fc5e11ba79c7472a22e): 
- Script to remove "broken" movements and transfers
- DB constraint

app/assets/javascripts/give_time.js.coffee
```diff
@@ -5,8 +5,13 @@ jQuery.validator.addMethod "either-hours-minutes-informed", ((value, element) ->
...
     submitHandler: (form) ->
-      $(" #transfer_amount ").val($(" #transfer_hours ").val() * 3600 + $(" #transfer_minutes ").val() * 60)
-      form.submit()
+      amount = $("#transfer_hours").val() * 3600 + $("#transfer_minutes").val() * 60
+      $("#transfer_amount").val(amount)
+
+      if amount > 0
+        form.submit()
+      else
+        $(form).find('input:submit').after(' <label class="error">Time must be greater than 0</label>')
```
enricostano: Shouldn't we i18n this?

markets:
Pushed i18n here https://github.com/coopdevs/timeoverflow/pull/421/commits/758f3dbb6d2040c80d4c0215f065aac2c70254a3 and rebased from develop.

config/locales/en.yml
```diff
@@ -494,6 +494,8 @@ en:
       minute:
         one: "%{count} minute"
         other: "%{count} minutes"
+    new:
+      error_amount: Time must be greater than 0
```
enricostano: Can be negative also?

sseerrggii: Can not

enricostano:
@sauloperez would you mind to finish your review here? :kissing_heart:

enricostano:
Deployed to https://staging.timeoverflow.org

cc/ @sseerrggii

sseerrggii:
Tested! :ok_hand: 

@markets can you rebase it? I have no courage :worried:

markets:
@sseerrggii rebased with current develop and push forced!

Issue #318, opened 2018-02-01
Remove N+1 from shared/_post view


sauloperez:
The line below is responsible for a N+1:

https://github.com/coopdevs/timeoverflow/blob/94a4ad8baa4e947f80d2cfef8c1bcf1ec287b95b/app/views/shared/_post.html.erb#L34

Due to:

https://github.com/coopdevs/timeoverflow/blob/94a4ad8baa4e947f80d2cfef8c1bcf1ec287b95b/app/models/post.rb#L122-L124

That method is going introduce N+1 every time is used, as it is very likely that is going to be used as part of a larger query.

We should rely on eager loading the `members` association when needed instead.

gpks:
I would love to try to fix this :)

sauloperez:
Hi @gpks ! please go ahead. Keep in mind that the app has changed considerably recently so make sure you check the N+1 still exists first. Sorry for any inconvenience.

enricostano:
@sauloperez can we close this issue? It seems that the N+1 is not there anymore. Right?

sauloperez:
It seems so. Let's close it

PR #435, opened 2018-09-25
Removed offers dashboard

mllocs:

PR #440, opened 2018-10-15
Remove n+1 problem

gpks:
I was trying to solve this https://github.com/coopdevs/timeoverflow/issues/318 but I couldn't find a problem, instead I've found other place with n+1 query, so I've fixed it.


markets:
Hi 👋 @gpks and thanks a lot for contributing! We're about to remove this part from the app, see #435.

PR #441, opened 2018-10-17
Add docker compose support

viktorsmari:
Hi, new contributor here!

I use **Docker** a lot and wanted to help out this project by making it easier to start up.

In this PR start with:
   `docker-compose up`  

And setup the db (only once) with
  `docker-compose exec app rake db:setup`

### Files changed:

1. `config/database.yml`

This could be a separate issue but Database creation fails on non-docker on my Ubuntu 16 machine with 
``` Couldn't create database for {"adapter"=>"postgresql", "username"=>"postgres", "collation"=>"es_ES.UTF-8", "ctype"=>"es_ES.UTF-8", "template"=>"template0", "encoding"=>"utf8", "host"=>"localhost", "database"=>"timeoverflow_development"}
PG::WrongObjectType: ERROR:  invalid locale name: "es_ES.UTF-8"
: CREATE DATABASE "timeoverflow_test" ENCODING = 'utf8' LC_COLLATE = 'es_ES.UTF-8' LC_CTYPE = 'es_ES.UTF-8' TEMPLATE = "template0"
```
This fails both in docker and non-docker environments. I 'fixed' this by commenting out `ctype` and `collation` in `config/database.yml` - We can add that as an ENV?

2. `Gemfile` 
* Having this in the Gemfile: 
`ruby File.read('.ruby-version').strip` 
does not work with Docker,  I removed it in this PR. It was introduced in #341
* Rails 5.2 projects will have the `ruby 'version'`  directly https://github.com/rails/rails/pull/30016
* I think we should do the same and add the string directly. I can create a different PR for that

### Tests
Tests can be run with `docker-compose exec app rake`

Currently **6** tests fail:
   **2** fail with this [known issue](https://github.com/elastic/elasticsearch-rails/issues/756) and might be fixed by updating the elasticsearch gems:
```
    Elasticsearch::Transport::Transport::Errors::NotAcceptable:
       [406] {"error":"Content-Type header [application/x-www-form-urlencoded] is not supported","status":406}
```

**4** fail with:
```
     Selenium::WebDriver::Error::WebDriverError:
     unable to connect to chromedriver 127.0.0.1:9515
```

#### Elastic search
Which version of elasticsearch are we using in prod?
   * We can imitate that by telling `docker-compose.yml` which version we want.

#### Instructions
Should we add instructions to the wiki or to the README?
I think basic instructions can live in the main README

**Notes:**
This is partly based on an older incomplete pr #187

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM ruby:2.3.0
+
+# Install essential Linux packages
+RUN apt-get update -qq && apt-get install -y \
+    build-essential \
+    libpq-dev \
+    postgresql-client \
+    nodejs
+
+WORKDIR /app
+
+# Copy Gemfile and Gemfile.lock
+COPY Gemfile* /app/
+
+# Speed up nokogiri install
+ENV NOKOGIRI_USE_SYSTEM_LIBRARIES 1
+
+RUN bundle install
+
+# Copy the Rails application into place
+COPY . /app
+
+#CMD [ "rails", "server", "-b", "3000" ]
+#CMD [ "bundle", "exec", "puma" ]
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,5 @@
 source 'https://rubygems.org'
 
-ruby File.read('.ruby-version').strip
 
 gem 'rails', '~> 4.2'
 gem 'rails-i18n'
--- a/config/database.yml
+++ b/config/database.yml
@@ -11,6 +11,7 @@
 #
 
 defaults: &defaults
+  host: <%= ENV['DATABASE_HOST'] || "localhost" %>  # use localhost as fallback
   adapter: postgresql
   username: <%= ENV['DATABASE_USER'] %>  # default is null
   collation: 'es_ES.UTF-8'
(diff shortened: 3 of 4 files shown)


enricostano:
Hi @viktorsmari , welcome to TimeOverflow community and many thanks for your time!

To start a working development environment we use LXC containers + Ansible, please give a look at https://github.com/coopdevs/timeoverflow-provisioning repo for more info and instructions.

DISCLAIMER: we know that LXC is not compatible with Mac OS and that Docker would be an option, but the limited resources that we have don't allow us to maintain any development platform other than GNU/Linux.

I hope you'll understand (... and that you use GNU/Linux :trollface: :joy: ).

viktorsmari:
I use only GNU/Linux :tada: but I think we should help people to get started fast by providing these files.
I think this project will keep receiving a PR like this if we don't provide it, because Docker is becoming a standard.

In my experience, when a project is Dockerized, it does not require maintenance.
We should also add a Disclaimer that '**the core devs do not use Docker, use at your own risk**' or something similar. What do you think about that @enricostano ?

enricostano:
We had this discussion a couple of times in the past, we also had unmantained `Dockerfile` and `Vagrantfile`... and the core team decided to only maintain things that it uses daily.

There is also another reasoning behind this decision, we use the same strategy to provision all the stages: development, staging and (coming soon) production. Adding a new way to run a development environment on GNU/Linux would compromise this strategy. I hope you can understand.

In any case you're more than welcome to join us :muscle:

Issue #454, opened 2018-12-12
Unexpected session expiration in secondary device 


sseerrggii:
I have been investigating why the session closes in the mobile app (webview). 

After a bit of headache trying to find the pattern, I think I understood what is going on.

The problem happens when user check "remember me" on login: this is the default behaviour on mobile app. In DB is stored as `REMEMBER CREATED AT` date

After that if this user login on desktop and close the session on desktop `REMEMBER CREATED AT` field is reset to null, and the session on mobile will expire in 1 hour (devise config.timeout_in)

So is there a way to avoid this behaviour? We want to keep `REMEMBER CREATED AT` on app session.

I found this config in `/config/initializers/devise.rb`, I'm not sure if its the solution: 

```
# Set this configuration to false if you want /users/sign_out to sign out
# only the current scope. By default, Devise signs out all scopes.
# config.sign_out_all_scopes = true
```

## Extra
We want to persist the session on mobile app, the remembereable option is for 4.weeks so we need to update `REMEMBER CREATED AT` the next time the user interact with the app, I think this can be done also in  `/config/initializers/devise.rb`

```
# If true, extends the user's remember period when remembered via cookie.
config.extend_remember_period
```

What do you think? @markets @mllocs @sauloperez @enricostano

sauloperez:
I don't know how that column actually looks like in the DB and I'm not an expert on Devise myself but what you suggest of sign_out_all_scopes sounds good.

sseerrggii:
Sadly don't work as expected :disappointed: 

`config.sign_out_all_scopes` is true but when I sign out session started without 'remember me' the `REMEMBER CREATED AT` is reset

Also I don't see that `config.extend_remember_period` refresh date on `REMEMBER CREATED AT`

Actually I'm not sure what is the behaviour expected of changing this variables. I didn't find it in Devise documentation  https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Rememberable

sauloperez:
Then we might need to check Devise's codebase

mperezv:
I think this matches with the behaviour described on this Devise's issue: https://github.com/heartcombo/devise/issues/5212

mperezv:
Also, I think we could have tried to set `expire_all_remember_me` to `false` but I found this other Devise's issue and it seems it wouldn't work: https://github.com/heartcombo/devise/issues/5027

PR #456, opened 2019-01-02
change some Devise params

sseerrggii:
As explained on #454 we will try to change this Devise parameters.

I will need to test it on staging and staging app

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -130,7 +130,7 @@ Devise.setup do |config|
   config.remember_for = 4.weeks
 
   # If true, extends the user's remember period when remembered via cookie.
-  config.extend_remember_period = false
+  config.extend_remember_period = true
 
   # Options to be passed to the created cookie. For instance, you can set
   # :secure => true in order to force SSL only cookies.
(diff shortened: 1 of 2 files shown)


config/initializers/devise.rb
```diff
@@ -130,7 +130,7 @@
   config.remember_for = 4.weeks
 
   # If true, extends the user's remember period when remembered via cookie.
```
sauloperez: Do we remember via cookie?

sseerrggii: Yes

markets:
@sseerrggii @sauloperez I pushed this change in order to fix the CI: https://github.com/coopdevs/timeoverflow/pull/456/commits/dd0e4875e45f2fadda22d97c8078b5179c205711. 

We should re-test this thing :pray: #432

enricostano:
Deployed to staging. cc/ @sseerrggii

sseerrggii:
Sadly don't work as expected 

`config.sign_out_all_scopes` is true but when I sign out session started without 'remember me' the `REMEMBER CREATED AT` is reset

Also I don't see that `config.extend_remember_period` refresh date on `REMEMBER CREATED AT`

Actually I'm not sure what is the behaviour expected of changing this variables. I didn't find it in Devise documentation https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Rememberable

enricostano:
@sseerrggii what happened? Not needed any more?

sseerrggii:
Sorry I didn't explain why :stuck_out_tongue_winking_eye: 
This PR is old and I want to do a new PR with a diferent approach to the problem

PR #464, opened 2019-02-20
Small controllers refactor

markets:
## WAT?
Couple of minor cleanups:

- removed unused store_location: #249 has been closed, so there is no point of maintaining a `session[:previous_url]` right now

- moved `switch_lang` action to `ApplicationController` + added spec: => :scissors: `GlobalController` that only handles this action, so I think it's fine (and simpler) to move it to main/app controller 

## TESTING
- We just need to test the language switcher


sseerrggii:
Tested :green_heart: 

language switcher works as usual

PR #465, opened 2019-02-21
Switch from elasticsearch to PostgreSQL text search

enricostano:
### WAT
Elasticsearch is a dependency difficult to maintain and having to maintain a mirrored copy of the resources is also difficult, as you can see looking at the code deleted by this PR.

We already have PostgreSQL as a dependency and since our needs are quite simple we'll use the PostgreSQL built in [Full Text Search](https://www.postgresql.org/docs/current/textsearch-intro.html).

The gem that I'm proposing is https://github.com/Casecommons/pg_search

### Depends on
https://github.com/coopdevs/timeoverflow-provisioning/pull/62

### Extra
Closes #293

PR #467, opened 2019-03-07
Enable partial word searches in pg_search

markets:
:eyes: :point_right: Target branch: `feature/switch-from-elastic-to-pg-search-la-buena` (#465)

---

Enable partial word searches in pg_search.
 
Doc: https://github.com/Casecommons/pg_search#prefix-postgresql-84-and-newer-only

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -3,8 +3,11 @@ class Post < ActiveRecord::Base
   include PgSearch
 
   pg_search_scope :search_by_query,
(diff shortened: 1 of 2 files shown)


enricostano:
@sseerrggii what dou you think? would you like to test it?

sseerrggii:
Tested :+1: 

I think now Post search have the same behaviour than current production environment with Elastic Search :tada:

PR #489, opened 2019-04-08
Re-activate Gender field + chart

### LINKED ISSUE #482: Reactivate "gender" field on User
As we discussed on https://community.coopdevs.org/t/reactivar-genero-en-perfil-de-usuario-estadisticas/881/4

We want to reactive gender field on User table. 

We want to add new question on user create/edit form 

**Which gender do you identify with?**

And use a select for te answer: 
* (empty) (by default)
* Women
* Men
* Other
* I prefer not to answer

This question is only for diversity community analysis purposes, so we won't show it in any view. 

We want to add new graph for admins on Statistics -> Demographics data below age graph, same kind of age graph

markets:
Closes #482 

Since I used part of the old code, most translations are still valid, we just need to translate new options: "Other" and "I prefer not to answer".

Captura de pantalla 2019-04-08 a las 23 59 03

### DIFF
--- a/app/controllers/statistics_controller.rb
+++ b/app/controllers/statistics_controller.rb
@@ -74,6 +74,7 @@ class StatisticsController < ApplicationController
   def statistics_demographics
     @members = current_organization.members
     @age_counts = age_counts
+    @gender_counts = gender_counts
   end
 
   def statistics_last_login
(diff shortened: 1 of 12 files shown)


sseerrggii:
Wooow :clap: :clap: :clap:
 I will translate it in release process :)

Tested on staging :cactus: 

giphy (6)

:heart: @markets

PR #498, opened 2019-05-16
 Pre-release translations

sseerrggii:


sseerrggii:
Seems like localeapp use 'other' as a function [for pluralize](https://github.com/Locale/localeapp/wiki/Create-Pluralised-Translations#creating-a-pluralized-translation) and for this reason we can't add it to localeapp 

FireShot Capture 225 - coopdevs_timeoverflow - Locale - app localeapp com

I'll commit a solution (change other for others) in this PR, is it ok?

config/locales/ca.yml
```diff
@@ -345,6 +349,25 @@ ca:
...
       text_donation: L'associació ADBdT ofereix TimeOverflow de forma gratuïta als Bancs de Temps. Amb una donació de %{href} pots ajudar a mantenir i millorar la plataforma.
       text_donation_link: 1€ al mes
+  multi_transfers:
+    multi_transfers: 'Transferències múltiples '
+    step:
+      confirm:
+        description: Confirma els canvis
+        type_of_transfer: Tipus de transferència
+      select_type:
+        description: Selecciona  el tipus de transferència
```
enricostano: There are 2 spaces between `Selecciona` and `el`, right?

config/locales/ca.yml
```diff
@@ -345,6 +349,25 @@ ca:
       text: "%{organization_name} en"
       text_donation: L'associació ADBdT ofereix TimeOverflow de forma gratuïta als Bancs de Temps. Amb una donació de %{href} pots ajudar a mantenir i millorar la plataforma.
       text_donation_link: 1€ al mes
+  multi_transfers:
+    multi_transfers: 'Transferències múltiples '
```
enricostano: Do we need that extra space at the end of the line? I see that in many copies.

config/locales/ca.yml
sseerrggii: OCD :+1: I will fix it

config/locales/ca.yml
sseerrggii: Not need it, i'll check it

sseerrggii: There is not spaces in localeapp, there is somthing that adds localeapp :thinking: also I'm not sure what rule localeapp use to add quotes...

enricostano: :see_no_evil:

app/models/user.rb
```diff
@@ -14,7 +14,7 @@ class User < ActiveRecord::Base
   GENDERS = %w(
     female
     male
-    other
+    others
```
sseerrggii: @markets what do you think to change that? I try to solve this problem https://github.com/coopdevs/timeoverflow/pull/498#issuecomment-492976580

I think is enough to change it here and in ymls, I tested locally and will do it again in staging

markets: Yep, I think this is enough 👍

sseerrggii:
Tested :heavy_check_mark:

PR #503, opened 2019-05-25
update to Ruby 2.6

markets:
:warning: This PR points to `feature/ubuntu-bionic` branch (#502)

Let's move to latest stable release :shipit:


markets:
😂 not really fixing it last night, but got the Travis 🛑 email and I remembered that just recently I was fixing those problems in some gems...

IMHO yes, it's worth it! It's an "small" patch that not affects runtime, only controller tests. Since we are going to do a Ruby upgrade and test/review all app... why not? It doesn't force to upgrade Rails right now either, and anyway, we should start to think in a Rails upgrade, our current version 4.2.Z is almost out of "Severe Security Issues" (Rails 6 is in rc already).

enricostano:
> It doesn't force to upgrade Rails right now either, and anyway, we should start to think in a Rails upgrade, our current version 4.2.Z is almost out of "Severe Security Issues" (Rails 6 is in rc already).

That's my point, we have to upgrade Rails ASAP. And that will bring latest Ruby with it.

> It's an "small" patch that not affects runtime, only controller tests. Since we are going to do a Ruby upgrade and test/review all app... why not?

OK then, let's do this! Thanks for your help as always 💪❤️

markets:
I am a volunteer to work on the Rails upgrade :hand: 

Thanks @enricostano! :ok_hand:

enricostano:
> I am a volunteer to work on the Rails upgrade

Cool! Perfect, let's plan the work needed 💪

PR #519, opened 2019-10-07
Remove plpgsql PostgreSQL extension

sauloperez:
It was added in the very beginning of the project in commit `e37eb7d` but it doesn't seem to be used anywhere. Leftovers of an old copy-paste?

cc @markets @enricostano


markets:
No idea what is that and why it was added (the commit doesn't provide such info https://github.com/coopdevs/timeoverflow/commit/e37eb7d).. so 👍 from my side

sseerrggii:
Tested :green_heart:

PR #520, opened 2019-10-07
Remove pg_trgm PostgreSQL extension

sauloperez:
No need to install this extension used by full-text search until we don't replace ElasticSearch with it.

Still, it was added in `f514c35` in the early days (2013) for user search but it doesn't seem to be used.

You @enricostano and @markets might now better.


sseerrggii:
Tested :heavy_check_mark:

PR #535, opened 2020-01-21
Remove next environment

sauloperez:
It was used while migration the production server and since this one and production configs are almost exact copies (just rails_env differs) it's no longer needed.

And this brings simplicity back to this repo.

The most noticeable change is the removal of `url` from production's database config as DATABASE_URL got removed in https://github.com/coopdevs/timeoverflow-provisioning/pull/21 two years ago and it's true in production:

```
timeoverflow@timeoverflow-production:~$ env | grep DATABASE
DATABASE_NAME=xxx
DATABASE_USER=xxx
DATABASE_PASSWORD=xxx
```

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -33,7 +33,3 @@ staging:
 production:
   <<: *defaults
   database: <%= ENV.fetch('DATABASE_NAME', 'timeoverflow_production') %>
-
-next:
-  <<: *defaults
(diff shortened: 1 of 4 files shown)


sauloperez:
Rebased with the changes introduced in https://github.com/coopdevs/timeoverflow/pull/536. No behavior changes :ok_hand:

PR #545, opened 2020-02-14
Fix/capybara language

mperezv:
Set default.locale to english


mperezv:
I just realized that this isn't passing Travis tests, while it wass passing the local tests. I'm going to check it again.

PR #546, opened 2020-02-17
Do /tags view responsive

### LINKED ISSUE #541: Some words in /tags overlap on small screens
We need to fix the margins.
TimeOverflow_Overlap

mperezv:
Set .col to better values for the view being responsive. The current view on small screens is like this below:
tags-siendo-responsive

Fix #541


sauloperez:
You started this branch from `master` and that is why you brought 18th other commits that are not made by you. You should use `git rebase -i develop` and remove all of them but the last one, which is yours. Either @danypr92 or I can help you with that if you need to.

PR #401, opened 2018-08-15
move weekly digest email to sidekiq-cron and remove whenever

markets:
Closes #373 
Closes https://github.com/coopdevs/timeoverflow-provisioning/issues/24
Closes #544
Closes #548 

**EXTRA:**
- update more dev deps: rubocop, byebug, fabrication, capybara, webdrivers
- fixing CI (feature specs)

**INFRA CHANGES & TESTING**
👀 https://github.com/coopdevs/timeoverflow/pull/401#issuecomment-471716708

PR #548, opened 2020-02-17
Set default.locale to en

mperezv:
We were having problems with tests as much in local as in Travis because they weren't able to find the submit button on login, even though we were pretty sure of the button being there. At last, we found out that the page was printing in english and the test was looking for a spanish content of tag. So it wasn't able to find it.

In order of the test finding the submit button on login we set default.locale variable to english on  `config/environments/test.rb` to change the language in which the test looks for the button. 

It worked the other day in local, I have a witness (@sauloperez), but for some reason it's failing on Travis and local today. So, currently, I'm trying to figure out how to fix it for real.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -6,6 +6,8 @@ Rails.application.configure do
   # your test database is "scratch space" for the test suite and is wiped
   # and recreated between test runs. Don't rely on the data there!
   config.cache_classes = true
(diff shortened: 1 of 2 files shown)


sauloperez:
@markets you might know what's going on

markets:
👋 Taking a look in #401

PR #554, opened 2020-02-28
Move spec from controller to view

mperezv:
We moved the 'displays the organization page' spec from controller specs to view specs and wrote it as a proper view spec.

### DIFF
--- a/spec/controllers/organizations_controller_spec.rb
+++ b/spec/controllers/organizations_controller_spec.rb
@@ -5,16 +5,6 @@ RSpec.describe OrganizationsController do
   let(:member) { Fabricate(:member, organization: organization) }
   let(:user) { member.user }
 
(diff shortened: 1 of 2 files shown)


spec/controllers/organizations_controller_spec.rb
```diff
@@ -5,16 +5,6 @@
   let(:member) { Fabricate(:member, organization: organization) }
   let(:user) { member.user }
 
-  describe 'GET #show' do
-    it 'displays the organization page' do
-      get 'show', id: organization.id
-
-      expect(assigns(:organization)).to eq(organization)
-      expect(response.status).to eq(200)
-      expect(response.body).to include(organization.name)
```
sauloperez: I would only remove this line. The rest is actually testing the controller's behavior and we still want it.

PR #556, opened 2020-02-28
send weekly digest email in user's locale

### LINKED ISSUE #453: Send email notifications in user language
Users receives email notifications always in ES

We want to send notifications in language chosen by user (LOCALE field in user table)

http://community.coopdevs.org/t/eleccion-de-idioma-en-el-boletin-de-avisos-semanal/730

markets:
closes #453

### DIFF
--- a/app/jobs/organization_notifier_job.rb
+++ b/app/jobs/organization_notifier_job.rb
@@ -24,7 +24,11 @@ class OrganizationNotifierJob < ActiveJob::Base
   private
 
   def users_by_locale(organization)
-    organization.users.online_active.actives.
-      notifications.group_by { |u| u.locale || I18n.default_locale }
+    with_notifications = organization.users.online_active.actives.notifications
+    org_locales = with_notifications.pluck(:locale).uniq
+
+    org_locales.each_with_object({}) do |locale, hash|
+      hash[locale] = with_notifications.where(locale: locale)
+    end
   end
 end
--- a/app/jobs/organization_notifier_job.rb
+++ b/app/jobs/organization_notifier_job.rb
@@ -2,7 +2,7 @@
 #
 # Strategy: go throught all organizations and take latest active posts from last week
 # posted by active members. Send an email to all active and online members
-# with the email notifications enabled with those posts.
+# with the email notifications enabled with those posts. Group emails by user's locale.
 
 # Schedule defined in config/schedule.yml file.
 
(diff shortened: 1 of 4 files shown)


app/jobs/organization_notifier_job.rb
```diff
@@ -12,9 +12,19 @@ class OrganizationNotifierJob < ActiveJob::Base
...
+        end
       end
     end
   end
+
+  private
+
+  def users_by_locale(organization)
+    organization.users.online_active.actives.
+      notifications.group_by { |u| u.locale || I18n.default_locale }
```
sauloperez: We already have the locale column defaulting to `es` so we can ask PostgreSQL to do the grouping for us, right? I'm just a bit concerned about the little visibility we have over the data fetching we do from this background job. It's the kind of thing we might realize too late that it's killing the DB.

markets: ah yes! that was my initial idea too (avoid in-memory sorting), but I thought that column doesn't have a default... but yes, you're right, it has a default in the schema (maybe I have some broken records in local DB), so we can rely on that.

markets: @sauloperez I'm just trying to move the grouping thing to postgres, but if I'm not wrong we'll need anyway some Ruby and in-memory operations to mount a hash or some data-structure like that:

```ruby
{
  es: [array of users with ES],
  en: [array of users with EN],
  ca: [array of users with CA]
  ...
}
```

Another way to implemente this, would be 1 query for each available locale in the given organization.

sauloperez: I tried to a quick look at this last week and didn't get anything. The problem is that we don't want only the locales but also the users within each one and that's nothing we can get with `GROUP BY`. The only way would be with PostgreSQL window functions but it feels too much to stuff into ActiveRecord for this simple use case. A query per locale seems enough now.

markets: Yep agree, I see "feasible" 2 options:
- group_by: 1 query, but more memory/cpu consuming due to Ruby in-memory operations [current solution]
- multiple queries: 1 query x locale, but no in-memory sorting

markets: @sauloperez approach (2) pushed in 8f62ec6

markets:
New approach as discussed in https://github.com/coopdevs/timeoverflow/pull/556#discussion_r393142110 and improved code coverage 💪

app/jobs/organization_notifier_job.rb
```diff
@@ -12,9 +12,23 @@ class OrganizationNotifierJob < ActiveJob::Base
...
+  private
+
+  def users_by_locale(organization)
+    with_notifications = organization.users.online_active.actives.notifications
+    org_locales = with_notifications.pluck(:locale).uniq
+
+    org_locales.each_with_object({}) do |locale, hash|
+      hash[locale] = with_notifications.where(locale: locale)
+    end
+  end
```
sauloperez: :ok_hand:

sauloperez:
@sseerrggii this is ready for a quick round of testing.

sseerrggii:
Tested :green_heart: 

Supereasy to test with sidekiq @markets :wink:

PR #565, opened 2020-04-24
Upgrade Ransack from 1.8.6 to 1.8.10

sauloperez:


sauloperez:
I checked the CHANGELOG and the changes themselves and I think there's little risk here. Mostly staff for future Rails versions.

PR #568, opened 2020-10-06
Remove Elastic, 2nd attemp

### LINKED ISSUE #551: Remove Elasticsearch
Back to PostgreSQL full text search again (~> revert https://github.com/coopdevs/timeoverflow/commit/096fcce484ffd4f843955bcc2cd36627e26e43bc).

No memcached, no server crons, no Elastic, no Angular ... such a world, only Ruby, PostgreSQL and Redis 👌 🔝 

@enricostano @sauloperez just adding an issue to track the status of this.

### LINKED ISSUE #558: Replace db/schema.rb with structure.sql
## Description

The way we aim to migrate from ElasticSearch to PostgreSQL full-text search requires a function and a trigger, which is not supported by default Rails Ruby schema format.

This change should be a refactor. No behavior change at all. That's what we'll to confirm in production.

## What to do

We just need to bring the change done in https://github.com/coopdevs/timeoverflow/pull/465/ in a new branch that only changes the schema.

Some time ago I came across a blog that would be great to check when doing this: https://blog.appsignal.com/2020/01/15/the-pros-and-cons-of-using-structure-sql-in-your-ruby-on-rails-application.html

markets:
Reverts #513 (-> which reverted #465) 😺 

Closes #551
Closes #558


markets:
@enricostano @sseerrggii works fine for me locally, and tests are passing ✅  too both local and in Travis. Ready for staging!

db/migrate/20190523225323_enable_unaccent_extension.rb
```diff
@@ -0,0 +1,9 @@
+class EnableUnaccentExtension < ActiveRecord::Migration
+  def up
+    enable_extension "unaccent"
```
sauloperez, line 3: :see_no_evil: this requires SUPERUSER privileges. I think I'll just do what we did with https://github.com/coopdevs/timeoverflow-provisioning/blob/316c6dfffef7b9a0eafd7737a40eb30c81b8e3c4/roles/database/tasks/main.yml#L56-L61

sauloperez: Done in https://github.com/coopdevs/timeoverflow-provisioning/pull/182 already.

sauloperez:
Tested and approved by @sseerrggii

PR #571, opened 2020-11-06
add ID to report popup so we can do more analisis

dataf3l:


sseerrggii:
Thanks @dataf3l 

This PR comes from our conversation, now the report looks like 

Screenshot_2020-11-06 INFORME

Banco de Tiempo de Bogotà requests to add ID in order to identify the Post, I think is an easy and good improvement!

We will also have to think about improving reports feature in the future to made it more flexible

markets (approved):
Hi 👋🏼 just a minor detail... would be preferable to avoid inline styles, maybe there is any Bootstrap (v3) helper to accomplish the same (or similar) visualization, IMHO it's worth to take a look.

Thanks!

dataf3l:
Ok, I’ll take a look

markets:
💡 maybe `col-md-1` would do the trick

**PS** since now we have 3 `td`, we should add a new `th` too

markets:
Moved to 👉🏼  #581

PR #587, opened 2021-01-08
make login box more responsive

### LINKED ISSUE #570: Login responsive
we have problems in login form on mobile when keyboard is needed

Screenshot_20201105-131921

markets:
Closes #570 
Now the box is not vertical aligned, but at least we allow scroll, so in case the mobile keyboard shows up, we are still able to focus into the inputs.


sseerrggii:
For me is good solution! :ok_hand:

PR #589, opened 2021-01-08
Rails 6.1

markets:
🎺 🎺  Living on the edge 🎺 🎺 

Now that we are on Rails 6, let's try to be as much updated as we can.

**NOTE** After reviewed the release changelog and played a bit locally (tests are also passing ✔️ ), this seems safe to merge

PR #590, opened 2021-01-09
add empty config/storage.yml to avoid app crashes on boot (eager_load)

markets:
Fixes https://github.com/coopdevs/timeoverflow/pull/589#issuecomment-757102480 @sseerrggii 

We are not using ActiveStorage yet, but last Rails version seems to require this file to exists (only noticeable in "eager load" environments).


sseerrggii:
IT works @markets :heavy_check_mark:

PR #592, opened 2021-02-21
keep org filter (scope) in public search

### LINKED ISSUE #508: organization filter is lost when filtering
In the public url for offers & inquires, i.e: 
https://www.timeoverflow.org/inquiries?org=266
If we make a search, the org parameter is lost, showing the inquires of the other time banks: https://www.timeoverflow.org/inquiries?q=x

markets:
Closes #508

Ideally, we should implement the "scope everything" project, we all agree... but that will take time, meanwhile this is probably worth it.

**Extras**
- updated rails ⬆️ , patch version (I reviewed the diff and the changelogs, it's safe to upgrade)
- refactored Reports CSV/PDF code, much more DRY 👌🏼  (I tested this and works fine, but @sseerrggii will take a second look 🙏🏼  please)


sseerrggii:
Tested! :heavy_check_mark: 

- [x] Searching loggedout with and without org
- [x] Export reports to CSV & PDF

Thanks @markets :clap:

PR #608, opened 2021-03-18
Do not allow login to inactive members -- solved

### LINKED ISSUE #472: Do not allow login to inactive members
Members can be disabled by admin as first step before to remove them from organisation. Member have "active" field  https://github.com/coopdevs/timeoverflow/blob/develop/db/schema.rb#L100

Members where active="false" should not be allowed to log this organisation. Notice that user can be member to different organisations.

franpb14:
In this branch I focused on solving the problem related to the issue #472. The solution that I suggest is the following: an user trying to access an organization in which he/she was deactivated will be redirected to a view that shows a warning message as well as a table with a list of the organzations (and all the information related to those organizations). The behavior of offers and inquiries is the same as for that of non-authenticated users, with the exception of those offers/inquiries coming from an organization in which they are active. 

I additionnaly changed the "set_current_organization" so those users who had been deactivated in their first organization do not have to see the error message every time they log in.

Closes #472


markets:
Hi @franpb14 👋🏼  and very welcome to TimeOverflow! 😄 

As this is your 1st pull request here, let me add some general  comments with tips, good practices, style ...

- First of all, this branch seems a bit outdated with develop. According to the `branches` page: https://github.com/coopdevs/timeoverflow/branches you are 20 commits behind `develop` branch (our base branch).
- Instead of mixing branches with bugfixes and features, we should focus on small problems, since they simplify the diff (-> the review), the testing, and they easier to reason about. So, to solve #472 we don't need all these changes and behavior changes. Let's focus on small pieces 🙏🏼 ! 
- Code style: there are some issues about indentation (we use 2 spaces for HTML too), empty lines, comments in spanish... that should be addressed first. You can use Rubocop locally to be sure your new code matches our current code style.
- In general, having logic or sql queries in the views, is a bad practice and we should try to avoid it always (it makes testing/maintenance really hard).
- We should also take care with code coverage, according to CodeClimate we are losing 7% of coverage in this branch :(

To move forward with this branch, IMO, we should just fix the problem described in #472. Let's start with small changes, step by step, it will be easier for all the team to give feedback, perform the code review, write all necessary tests...

Thanks a lot for working on this and welcome to TimeOverflow Fran (and sorry for this long comment 😜 )

franpb14:
Thanks @markets for the review, I expected something like this, but I have some doubts: 
- To finish with this issue, I need to make a branch to fix it, another one to test it and another one to change the behavior, right?
- I don't understand why the coverage dropped because I tested my methods and, obviusly, I did not delete any tests. Maybe it's the outdated?
- When I start a larger task, if I don't test it on the same branch, the coverage will go down.
I'm looking forward to make some useful code for timeoverflow :)

markets:
> To finish with this issue, I need to make a branch to fix it, another one to test it and another one to change the behavior, right?

No, no, tests and code should be in the same branch, but mixing small fixes with behavior changes (which can have more implications) it's a bad idea in general. Small branches are the best! In that case, if we want to avoid the login for non-active users, we can just (for example) redirect back with a flash message "You can't login to this organization", with no need to implement new pages and workflows.

> I don't understand why the coverage dropped because I tested my methods and, obviusly, I did not delete any tests. Maybe it's the outdated?

Yep, this is right, probably it's just a matter of updating the branch.

---

Best,
Marc

franpb14:
Maybe I'm wrong but I think your proposed solution will create a critical bug. Without change the behaviour of login or without the new page, the users who were deactivated from the first organisation will not be able to login even if they are active in another organisation.

markets:
Ah got it! you're right about that point 👍🏼, now I understand better why you took this approach 😄 If @sseerrggii (the product owner) likes this, we can focus only on the "technical" aspects.

franpb14:
In the video we can see the changes.
First, the user logs in and sees the error because he was deactivated from the first organisation, he can see the bank information with a modal, and then we can see that the behaviour for offers/demands is the same as for unidentified and finally we can see that he can log in from the view and what happens when he tries to switch to an organisation from which he was deactivated.

https://user-images.githubusercontent.com/70280187/112175074-38240600-8bf7-11eb-9b43-871da31a6e19.mp4

I saw that travis failed but locally the tests run correctly and the failures in the log on Travis are different for each abuild. Why?

markets:
Hi @franpb14 that seems a flaky test, the PR build is passing. I tried to restart the build in Travis, but I got an error:

> Oh no! An error occurred. The build could not be restarted.

So, no worries about that point, the build is also passing for me locally. Probably it's an order dependent test (which means we have an issue in our specs 😄 ). In `rspec`, the order can be reproduced the via the `--seed` argument: `rspec --seed 1234`. The seed is printed on each run, in the terminal ("Randomized with seed 1234").

I'd add some extra comments/feedback but now seems much better in general 👌🏼  Thanks for reviewing the comments.

app/controllers/application_controller.rb
```diff
@@ -123,4 +123,11 @@ def user_not_authorized
   def resource_not_found
     render 'errors/not_found', status: 404
   end
+  
+  def member_should_be_active
+    if !current_member.active
+      flash[:error] = 'Your account have been deactivated from this timebank'
```
markets, line 129: we'll need i18n here

app/controllers/application_controller.rb
```diff
@@ -123,4 +123,11 @@ def user_not_authorized
...
+      redirect_to '/select_organization'
```
markets, line 130: better to use the route helpers generated by Rails, so we avoid hardcoding strings

app/controllers/organizations_controller.rb
```diff
@@ -49,6 +49,14 @@ def set_current
     redirect_to root_path
   end
 
+  def select_organization
+    @members = Member.where(user_id: current_user.id)
+    @organizations = Array.new
+    @members.each do |member|
+      @organizations << Organization.find(member.organization_id)
+    end
```
markets, line 57: I think we can replace this entire code block by `@organizations = current_user.organizations`

app/controllers/posts_controller.rb
```diff
@@ -5,8 +5,7 @@ class PostsController <  ApplicationController
 
   def index
     context = model.active.of_active_members
-
-    if current_organization.present?
+    if current_organization.present? && current_member.active?
```
markets, line 8: mmm 🤔  not sure about this one, posts index/show (read access) pages are public, so maybe we can avoid this check?

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
+<div class='table-responsive'>
+  <table class='table table-hover table-condensed users panel'>
```
markets, line 2: are the classes `users panel` doing something special? If not, I'd propose to remove them

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+    <thead>
+      <tr>
+        <th>
+          <%= t 'activerecord.attributes.organization.name'%>
```
markets, line 6: `' %>` note the space before the closing erb tag

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+        </th>
+        <th>         
+          <%= t 'global.Information'%>
```
markets, line 9: `global.information` lowcase

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+            <td><a href='' data-toggle='modal' data-target='#dialogo<%= index %>'> <%= t 'global.see'%></a></td>
+            <td>
+              <% if current_user.active?(organization) %> 
+                <%= link_to set_current_organization_path(organization), method: :post do %>
+                <%= t 'application.landing.button' %>
+              <% end %>
+              <% else %>
+                <%= t 'activerecord.attributes.user.deactivated'%>
+              <% end %>
+            </td>
```
markets, line 29: this `td`s are bad indented

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+        <th>         
+          <%= t 'global.enter_to_timebank'%>
+        </th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @organizations.each_with_index do |organization, index| %>
+        <tr>
+            <td><%= organization.name %></td>
+            <td><a href='' data-toggle='modal' data-target='#dialogo<%= index %>'> <%= t 'global.see'%></a></td>
```
markets, line 20: let's try to avoid non-english words for code 🙏🏼

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+            <td>
+              <% if current_user.active?(organization) %> 
+                <%= link_to set_current_organization_path(organization), method: :post do %>
+                <%= t 'application.landing.button' %>
+              <% end %>
```
markets, line 25: this line and the above line, need indentation: 

```erb
<%= link_to set_current_organization_path(organization), method: :post do %>
  <%= t 'application.landing.button' %>
<% end %>
```

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+              <div class='modal-header'>
+                <button type='button' class='close' data-dismiss='modal'>X</button>
+                <h4 class='modal-title'><%= organization.name %></h4>
+              </div>
+              <div class='modal-body'>
+                <strong><%= t 'activerecord.attributes.organization.address' %>:</strong> 
+                <%= organization.address %> <%= organization.neighborhood %> <%= organization.city %>
+                <br>
+                <strong><%= t 'activerecord.attributes.organization.phone' %>:</strong> 
+                  <%= organization.phone %>
```
markets, line 43: 👀  indentation

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+                <br>
+                <strong><%= t 'activerecord.attributes.organization.email '%>:</strong> 
+                  <%= organization.email %>
```
markets, line 46: 👀  indentation

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+                <br>
+                <strong>Web:</strong> 
+                  <a href=<%= organization.web %>><%= organization.web %></a>
```
markets, line 49: 👀  indentation

app/views/organizations/select_organization.html.erb
```diff
@@ -0,0 +1,63 @@
...
+              </div>
+              <div class='modal-footer'>
+                  <button type='button' class='btn btn-secondary' data-dismiss='modal'>Close</button>
```
markets, line 52: 👀 indentation

app/views/shared/_post.html.erb
```diff
@@ -32,7 +32,7 @@
         <% end %>
       </div>
     </div>
-    <% if current_user && current_organization == post.organization %>
+    <% if current_user && current_organization == post.organization && current_user.active?(current_organization) %>
```
markets, line 35: similar to: https://github.com/coopdevs/timeoverflow/pull/608/files#r599927728

spec/controllers/users_controller_spec.rb
```diff
@@ -96,6 +103,15 @@
       end
     end
 
+    context 'with a member deactivate' do
+      it 'is redirected to select organization' do
+        login(deactivate_user)
+        get :index
+
+        assert_routing('/select_organization', controller: 'organizations', action: 'select_organization')
```
markets, line 111: Would be nice to use the `redirect_to` matcher here: https://relishapp.com/rspec/rspec-rails/docs/matchers/redirect-to-matcher

You can find some examples in our suite too.

app/controllers/posts_controller.rb
franpb14: With this check you show all offers imitating the behaviour you would have if you were not logged in. Without the check, you only see the current timebank's postings.

app/views/organizations/select_organization.html.erb
franpb14: the users class does nothing but the panel class format the table.

markets: ok, let's remove the `users` class then

app/controllers/posts_controller.rb
markets: mmm ok I understand your change, but still not sure how we should handle this, maybe we should discuss it with @sseerrggii too. Maybe we can just hide the "navbar", if you are not active in this organization and set current_organization to nil. So, we can save this logic to be spread around the code.

sseerrggii: Yes, we have to try to limit the logic at the beginning of the session, avoiding to complicate this functionality further.

To my understanding there are two possible cases
1. The user is an active user of some organizations: When entering, access the first organization in which he is an active member. (In the menu that allows you to change the organization, you will also have to go through this logic before changing to a new organization, also don't show on the menu the organizations where are you an inactive member)
  
2. The user is not an active member of any organization, I think we can prevent him from logging in and show him a public page with the information that he is not an active member of any organization?

franpb14: then I need to change the current login behaviour and the menu to change the organisation, right?

sseerrggii: No need @franpb14 , it was only a proposal, I saw the behaviour https://github.com/coopdevs/timeoverflow/pull/608#issuecomment-805019448 and it's ok for me :+1:  I think we need only to avoid assign current organization where you are an inactive member, what do you think?

franpb14:
Thank you very much for taking your time correcting it, I really appreciate it. I did a new push implementing feedback but the current_organization can't be nil easily because in "application_controller" the organisation is set for each action so that when the user enters, for example, /offers  current_organisatio would be set again. After talking to @sseerrggii we decided not to do that for this issue.

markets:
@franpb14 @sseerrggii I pushed one commit with some minor tweaks (fa5c885446e02969b40caf87f31d72be587ba0cd). This is ready for testing from my side ✔️  Good job @franpb14 👍🏼 

ℹ️  @sseerrggii we'll need to add 4 new keys in Locale

markets (approved):

sseerrggii:
Works fine! Thanks @franpb14  and @markets :muscle:

PR #612, opened 2021-03-27
Delete org creation ability from public part

markets:
Superadmins can do it easily in the `Admin` section, so I propose to remove it from the public part, less is more!

**Extras**
- **styles** 🎨: 
  - fix footer to make it always fixed on bottom 
  - make members pages titles consistent across the rest of the app
- **admin**: allow to change membership "active" flag 🔧 

PS not sure why codeclimate is not happy about code coverage, I'm basically deleting code... I also checked it locally: develop 83.6% vs this branch 83.7% 🤷  CI is passing by the way, so let's ignore this coverage ❌


markets:
@sseerrggii ready for a quick test!

sseerrggii:
Tested!! :heavy_check_mark: 

Thanks @markets

Issue #607, opened 2021-03-17
Avoid resend transfer hours form 


sseerrggii:
As reported https://community.coopdevs.org/t/transferencias-duplicadas/1166

Seems to be a problem only reproducible on the mobile app. When you are in a paying hours form `/transfers/new?destination_account_id=6&id=4` if you click to submit botton several times the transfer will be saved several times

markets:
Closed by #614

PR #517, opened 2019-10-03
Remove LazyRecoverable and dalli (and memcached)

markets:
Closes #358
Closes #327

Added `data: { disable_with: '...' }` to the button to disable it while the request is running (should avoid fast double clicks).

PR #614, opened 2021-04-16
Transfers form: avoid duplicates when clicking several times

markets:
Hi @sseerrggii 👋🏼  this is an attempt to solve #607. Approach: use Rails native data-disable-with in the submit button, so when a user clicks, the button becomes disabled (similar as we did here: #517).

**EXTRA** 🎁 
Minor Admin improvements:
- Add some missing attributes/filters in index pages.
- Allow to properly create posts. Basically add an Organization selector in this form. In future iterations, we need better control for that (probably using nested selects via ajax + new validations), but at least now you will be able to create "valid" posts from the Admin.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -181,7 +181,7 @@ Style/SpecialGlobalVars:
   Enabled: false
 
 Style/StringLiterals:
-  EnforcedStyle: single_quotes
+  EnforcedStyle: double_quotes
 
 Style/VariableInterpolation:
   Enabled: false
--- a/app/admin/category.rb
+++ b/app/admin/category.rb
@@ -1,6 +1,5 @@
 ActiveAdmin.register Category do
   index do
-    selectable_column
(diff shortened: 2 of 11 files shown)


sseerrggii:
Tested!! 

Nice improvements @markets  :)

On production, admin selects always are too large and difficult to find the element you are looking for. I usually use organization selector and usually I do tasks about last organization created, so order orgs by ID on this select boxes is a good first step for me, Could you check if all org selects are ordered by ID in your next EXTRA features? Thanks :gift_heart:

markets:
👍🏼  yes! this is one of the "classic" gotchas of ActiveAdmin. I'll look for a proper solution in the next round! Since we already have the select2 library, maybe we can convert all these selects to "js selects" with the search integrated: https://select2.org/getting-started/basic-usage

sseerrggii:
Yes, this will be nice :)

PR #617, opened 2021-04-17
Admin: use select2

markets:
**WHAT?**

Basically this 👉🏼  https://github.com/coopdevs/timeoverflow/pull/614#issuecomment-821831122

**EXTRA**
- 💎  Add https://github.com/Shopify/bootsnap to speed up boot time (default gem since Rails 5.2)
- ✂️  `to-categories-dropdown.scss` => we already have Bootstrap helpers for that
- 🐛  fix some i18n keys (implicit scopes doesn't work well with partials and helpers, they are usually called in different places)
- 🔧  code style refactor in `Movement#other_side` method + test

PR #619, opened 2021-04-19
admin: fix post creation with tags 

markets:
In #617 I didn't realize that trying to create a post with tags (in the admin section) didn't work. Even before that PR (in master), didn't work either. So basically, now we'll able to send tags as string (comma separated) and create/update "complete" posts from the admin section 👌🏼 . I also added of couple missing filters :)

### DIFF
--- a/app/admin/post.rb
+++ b/app/admin/post.rb
@@ -2,6 +2,7 @@ ActiveAdmin.register Post do
   index do
     id_column
     column :class
+    column :is_group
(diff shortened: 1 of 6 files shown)


sseerrggii:
Is it ready to test @markets ?

markets:
Yes @sseerrggii

sseerrggii:
Tested :+1:

PR #622, opened 2021-04-22
Feat: searcher in "/organizations" working

### LINKED ISSUE #621: Feature: onboarding phase one
I am opening this Issue to implement a new functionality for those who are looking for a organisation. I've defined this Issue with the help of @sseerrggii, the first phase will be a listing with a searcher as you can see in the following mockups. 
First
Second

After clicking the user will see the organisation's information with a message to encourage the user to contact the organisation. 

The second phase could be to incorporate a map in addition to the list. The map could have both timeoverflow organisations and others.

franpb14:
The search engine works on address, name, city and neighborhood. The implementation is very similar to the offer searcher. Visitors will also see a message encouraging them to contact the organisation. You can see it working in the following video:

https://user-images.githubusercontent.com/70280187/115721881-be5e7400-a37e-11eb-896f-70a659dfcfde.mp4

[Problem with Travis](https://travis-ci.com/github/coopdevs/timeoverflow/builds/223883235): tests are running locally and all are fine. I tried the same order as in the failed build and it's ok locally as well.

- ¿Maybe the problem lies in the "migrate" needed to run the search engine in the organisation? Because it changed the "structure.sql".

closes #621.

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -2,7 +2,11 @@ class OrganizationsController < ApplicationController
   before_action :load_resource, only: [:show, :edit, :update, :set_current]
 
   def index
-    @organizations = Organization.all.page(params[:page]).per(25)
+    @organizations = if params[:q].present?
+                       Organization.search_by_query(params[:q]).page(params[:page]).per(25)
+                     else
+                       Organization.all.page(params[:page]).per(25)
+                     end
   end
 
   def show
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -2,11 +2,9 @@ class OrganizationsController < ApplicationController
   before_action :load_resource, only: [:show, :edit, :update, :set_current]
 
   def index
-    @organizations = if params[:q].present?
-                       Organization.search_by_query(params[:q]).page(params[:page]).per(25)
-                     else
-                       Organization.all.page(params[:page]).per(25)
-                     end
+    organizations  = Organization.all
+    organizations  = organizations.search_by_query(params[:q]) if params[:q].present?
+    @organizations = organizations.page(params[:page]).per(25)
   end
 
   def show
--- a/app/views/organizations/index.html.erb
+++ b/app/views/organizations/index.html.erb
@@ -1,6 +1,25 @@
 <h1>
   <%= Organization.model_name.human(count: :many) %>
 </h1>
+<div class="row">
+  <div class="col-md-6">
+    <form action="<%= organizations_path %>"
+          class="navbar-form navbar-left"
+          method="get"
+          role="search">
+      <div class="form-group">
+        <input  class="form-control"
+                name="q"
+                placeholder="<%= t "global.search_location" %>"
+                type="text"
+                value="<%= params[:q] %>">
+      </div>
+      <button class="btn btn-default" type="submit">
+        <%= t "global.search" %>
+      </button>
+    </form>
+  </div>
+</div>
 
 <div class="panel panel-default table-responsive">
   <div class="panel-body">
--- a/app/views/organizations/index.html.erb
+++ b/app/views/organizations/index.html.erb
@@ -8,7 +8,7 @@
           method="get"
           role="search">
       <div class="form-group">
-        <input  class="form-control"
+        <input class="form-control"
                 name="q"
                 placeholder="<%= t "global.search_location" %>"
                 type="text"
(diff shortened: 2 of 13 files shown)


markets:
nice one @franpb14! 👍🏼 

> ¿Maybe the problem lies in the "migrate" needed to run the search engine in the organization? Because it changed the "structure.sql"

I don't see any migration in this branch... 😄  When we integrated this feature in posts, we also added a tsv column, using a GIN index and DB triggers, for performance. But that seems a bit overkill and complex for this use case. So I'd try to avoid migrations and the tsv column.

app/controllers/organizations_controller.rb
```diff
@@ -2,7 +2,11 @@ class OrganizationsController < ApplicationController
   before_action :load_resource, only: [:show, :edit, :update, :set_current]
 
   def index
-    @organizations = Organization.all.page(params[:page]).per(25)
+    @organizations = if params[:q].present?
+                       Organization.search_by_query(params[:q]).page(params[:page]).per(25)
+                     else
+                       Organization.all.page(params[:page]).per(25)
+                     end
```
markets, line 9: what about something like?

```
organizations  = Organization.all
organizations  = organizations.search_by_query(params[:q]) if params[:q].present?
@organizations = organizations.page(params[:page]).per(25)
```

app/views/organizations/index.html.erb
```diff
@@ -1,6 +1,25 @@
...
   <%= Organization.model_name.human(count: :many) %>
 </h1>
+<div class="row">
+  <div class="col-md-6">
+    <form action="<%= organizations_path %>"
+          class="navbar-form navbar-left"
+          method="get"
+          role="search">
+      <div class="form-group">
+        <input  class="form-control"
```
markets, line 11: 👀  2 spaces

spec/controllers/organizations_controller_spec.rb
```diff
@@ -2,12 +2,55 @@
...
   describe 'GET #index' do
     it 'populates and array of organizations' do
       get :index
 
-      expect(assigns(:organizations)).to eq([organization])
+      expect(assigns(:organizations)).to eq([organization, second_organization])
+    end
+  end
+
+  describe 'GET #index (search)' do
```
markets, line 15: maybe we can move this as a `context` inside the `describe 'GET #index'` block, wdyt?

spec/controllers/organizations_controller_spec.rb
```diff
@@ -2,12 +2,55 @@
...
+    before do
+      second_organization.name = "Banco del tiempo Doe"
+      second_organization.city = "Sevilla"
+      second_organization.address = "Calle gloria"
+      second_organization.neighborhood = "La paz"
+      second_organization.save!
+      organization.neighborhood = "La paz"
+      organization.save!
+    end
```
markets, line 24: 🙏🏼  let's leave an empty line between each code block (before, it, describe, ...)

franpb14:
@markets My bad, reading the "pg_search" tutorial I misunderstood that it was mandatory to do a migrate. I've reverted "structure.sql".

spec/controllers/organizations_controller_spec.rb
franpb14: I followed the same structure as in "offers" but it is true that the "Get #index" of organizations is emptier so it could be fine.

app/controllers/organizations_controller.rb
franpb14: I think the efficiency of the algorithm you propose is worse, you always make a call to fetch all organisations even if the "important" query only fetches two, right?

markets: Not really, in ActiveRecord those queries are "lazy" evaluated, the first time a `to_a` or similar (`each` in the view) is invoked, so meanwhile you can keep chaining new scopes and/or filters.

spec/controllers/organizations_controller_spec.rb
markets: Yes, inside the block above, as a new `context` seems a better fit 👌🏼

app/controllers/organizations_controller.rb
franpb14: oh I didn't know, thanks for the explanation :)

markets: You can check the log to see where and how each query is exactly triggered.

markets (approved):
@franpb14 @sseerrggii ready for testing too! 👌🏼

sseerrggii:
:heart_eyes: 
Tested :heavy_check_mark: 

Thanks @franpb14 & @markets 

I will translate some copies and publish a new release :boom:

PR #626, opened 2021-04-26
feat: tags and postal code attributes

### LINKED ISSUE #623: Feature: search members by post code or tags
This feature is about being able to search for members of an organisation by postcode or tag (each organisation may choose its own tag policy or not use tags directly). With the postcode, large organisations can put their closest members in touch with each other.

Main points:

- [x]  Add tags (optional text) and postcode (optional string) fields and add them to the forms.
- [x] Make the current searcher work on the postcode and a dropdown with tags. (At the end you can see a mockup of this).
- [x] Autocomplete tags in forms.

The mockup:
MocupsBuscarUser

franpb14:
Postal code is related to users and tags is related to members (for those who belong to more than one organisation, they can have different tags depending on the organisation). 

Closes #623

https://user-images.githubusercontent.com/70280187/116120984-6e561900-a6c0-11eb-8080-8d08d92c43bc.mp4


sseerrggii:
:heart_eyes: 

Just to be sure @franpb14 tags are shared only on the same organization? This is how it works in Posts

franpb14:
Yes, tags are related to model is an attribute of the member model. so, for example, a user can have "tag1 and tag2" in one organisation and "otherTag and ultimaTag" in another. And if you want to see the tags of and if you want to see the tags of an organisation, just take the members of the organisation and extract the tags.

franpb14:
I'm working on second step and I have seen a bug, with this PR it is not possible to search for publications by tags. 
Screenshot from 2021-04-27 22-28-49

franpb14:
The problem was solved by indicating the table where to search for the tag in the query that was failing. [Here is the commit](https://github.com/coopdevs/timeoverflow/pull/626/commits/32f4ebe583ec571ab50c4f1c50a9cffd48048738) that solves it.

markets:
hi @franpb14 I think we can use this same branch and PR to add the other two points. The searcher and the autocomplete, seem quite "small" tasks (we have examples of both in current code base) and very tied to this code.

PS comment for latest commit 32f4ebe583ec571ab50c4f1c50a9cffd48048738: this is a concern (a mixin with steroids) to be injected in other classes, so hardcoding the a table name seems wrong to me, it should use the table of the injected AR class, we can do it by using the `table_name` method, something like: `where("? = ANY (#{table_name}.tags)", tag)`

franpb14:
All steps in #623 are complete. I have two videos, the first one is related to the searcher working on tags:

https://user-images.githubusercontent.com/70280187/116464729-9da98900-a86c-11eb-9746-cf423eb57c43.mp4

And the last video is related to tag autocompletion taking into account the members tags and how to show the tags in "/tags/alpha_grouped_index"

https://user-images.githubusercontent.com/70280187/116465157-258f9300-a86d-11eb-8208-b2973b5a88c4.mp4

sseerrggii:
OMG I love this feature, also I like really much that you add it on tags list page :clap: :clap: :clap:

app/assets/javascripts/application/tags.js
```diff
@@ -41,4 +45,26 @@ $(function() {
       }
     }
   });
+
+  $('#tags-js-members').select2({
```
markets, line 49: I'd delete this new initializer to make it DRYer. The only difference is the param you need to send to the backend, which can be pulled from a data-attribute, for example.

app/controllers/users_controller.rb
```diff
@@ -36,7 +43,7 @@ def create
     @user.setup_and_save_user
 
     if @user.persisted?
-      @user.tune_after_persisted(current_organization)
+      @user.tune_after_persisted(current_organization, get_tags)
```
markets, line 46: instead of passing around this tags variable in different methods, I'd try to use a different approach, some ideas:
- send the `tag_list` attribute (provided by teh Taggable concern) + nested attributes
- use a new method to update tags in all places the same way, ala `@user.add_tags(organization)`

app/controllers/users_controller.rb
```diff
@@ -46,9 +53,10 @@ def create
 
   def update
     @user = scoped_users.find(params[:id])
+    @member = @user.as_member_of(current_organization)
     authorize @user
 
-    if @user.update(user_params)
+    if @user.update(user_params) && @member.update(tags: get_tags)
```
markets, line 59: this probably can be fixed/changed by the proposal in https://github.com/coopdevs/timeoverflow/pull/626/files#r623289771

app/models/user.rb
```diff
@@ -102,8 +102,8 @@ def setup_and_save_user
     save
   end
 
-  def tune_after_persisted(organization)
-    add_to_organization organization
+  def tune_after_persisted(organization, tag_list = [])
+    add_to_organization organization, tag_list
```
markets, line 106: those changes passing these tags around the code, can be fixed by https://github.com/coopdevs/timeoverflow/pull/626/files#r623289771

app/views/users/index.html.erb
```diff
@@ -14,7 +14,7 @@
     <div class="col-md-12">
       <%= search_form_for(@search, class: "navbar-form navbar-left", url: users_path) do |f| %>
         <div class="form-group">
-          <%= f.search_field :user_username_or_user_email_or_member_uid_search_contains, class: "form-control" %>
+          <%= f.search_field :user_username_or_user_email_or_member_tags_or_member_uid_search_contains, class: "form-control" %>
```
markets, line 17: not sure how we can improve this :thinking:, but this name starts to be too much... maybe Ransack allows some kind "alias"?

spec/controllers/tags_controller_spec.rb
```diff
@@ -49,6 +60,15 @@
       })
     end
 
+    it "load member tags" do
+      get :alpha_grouped_index, params: { post_type: "user" }
+
+      expect(assigns(:alpha_tags)).to eq({
+          "H" => [["html", 2], ["html5", 1]],
+          "C" => [["css", 1]]
```
markets, line 68: ⚠️  indentation

markets (changes requested):
hello @franpb14 very nice improvements 👏🏼 👏🏼  I just completed the review and added some comments. Let me know what do you think?

db/structure.sql
```diff
@@ -91,7 +85,7 @@ SET default_tablespace = '';
 SET default_with_oids = false;
 
 --
--- Name: accounts; Type: TABLE; Schema: public; Owner: -
+-- Name: accounts; Type: TABLE; Schema: public; Owner: -; Tablespace:
```
markets, line 88: not sure what this `Tablespace:` is... but I'd remove it just in case 😄

app/views/users/index.html.erb
markets: found it! https://github.com/activerecord-hackery/ransack#ransack-aliases

we can create a new one for both members lists 👌🏼  so we don't need to write and maintain this large expression everywhere :)

app/controllers/users_controller.rb
franpb14: I'm not quite sure how to do the first one, so if they are just as efficient I'll do the second one.

franpb14:
Thank you very much @markets :). All changes are fine, I will make them between today and tomorrow.

franpb14:
@markets all the feedback is incorporated and I have also added member in case of failure of the forms.

markets:
@franpb14 pushed some refactoring 4d9d6a7

I'll take a second look tomorrow or during the weekend and we'll be ready for staging!

markets (approved):
@franpb14 @sseerrggii ready for staging! ✅ (i18n keys already added in Localeapp and pulled here)

sseerrggii:
Tested manually :heavy_check_mark:

PR #632, opened 2021-05-03
Changed the order of the buttons in the tag list + remove identity_document column

### LINKED ISSUE #630: Cleanup user fields
In Active Admin when I create a new user I see the field DNI (```identity_document```)

We don't use it and we can remove it from Active Admin and from User model. 

Also we introduced the field postal code (```postcode```) so it should be at active admin

### LINKED ISSUE #631: Tag List order
In Tag list we have the menu "Offers/Inquiries/Members" In order to continue with the same consistency as the main menu it should be arranged as "Members/Offers/Inquiries"

Screenshot_2021-05-03 Banc del Temps de Gràcia

franpb14:
It was also necessary to change a test because now if no parameter is passed it is of type "user".

Screenshot from 2021-05-03 11-58-41

Closes #631

**EXTRA**
- remove identity_document column from users (Closes #630)


markets (approved):
@franpb14 @sseerrggii this is ready for staging! 👌🏼

sseerrggii:
Tested!  :+1:

PR #639, opened 2021-05-16
Time exchanges can now be made between organisations

### LINKED ISSUE #638: Feature: transfers between members of different organisations
I will make the current "time transfer for this offer" button work even if it is an offer from another organisation. This will allow different organisations to exchange time with each other.

The button will work as follows when the exchange is between different organisations:
member A and his organisation give one hour and member B and his organisation get one hour. So now the sum of time of the members of organisation A will be -1 and of organisation B will be 1.

For this Issue the way to share an offer from another organisation will be via URL, then we will look at how to implement a search engine.

franpb14:
The feature is working exactly as shown in [this drawing](https://github.com/coopdevs/timeoverflow/issues/638#issuecomment-841066253). Here you can see it working in the following video and the full trace it would do:

https://user-images.githubusercontent.com/70280187/118401127-bedbe900-b664-11eb-8a57-abbe6b474654.mp4

If the source account is the organisation the first trace between first member and organisation would not be there but the rest is the same. When a member makes a donation it is the same, except that they cannot choose the source (as they do with current transfers).

For the implementation I have adapted the current transfers and therefore also had to change some tests (minor changes and specification of a new context).

closes #638.

### DIFF
--- a/app/models/transfer_factory.rb
+++ b/app/models/transfer_factory.rb
@@ -1,16 +1,22 @@
 class TransferFactory
-  def initialize(current_organization, current_user, offer_id, destination_account_id)
+  def initialize(current_organization, current_user, offer_id, destination_account_id,
+                 destination_organization_id)
     @current_organization = current_organization
     @current_user = current_user
     @offer_id = offer_id
     @destination_account_id = destination_account_id
+    @destination_organization_id = destination_organization_id.to_i
+  end
+
+  def destination_organization
+    Organization.find(@destination_organization_id)
   end
 
   # Returns the offer that is the subject of the transfer
   #
   # @return [Maybe<Offer>]
   def offer
-    current_organization.offers.find_by_id(offer_id)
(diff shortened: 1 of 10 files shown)


app/models/transfer_factory.rb
```diff
@@ -1,16 +1,22 @@
...
+  def destination_organization
+    Organization.find(@destination_organization_id)
   end
 
   # Returns the offer that is the subject of the transfer
   #
   # @return [Maybe<Offer>]
   def offer
-    current_organization.offers.find_by_id(offer_id)
+    destination_organization.offers.find_by_id(offer_id)
```
franpb14, line 19: I can solve the [codeclimate problem ](https://codeclimate.com/github/coopdevs/timeoverflow/pull/639) by changing this line to` Offer.find(offer_id)` and then changing the line of `destination_organization` to `offer.organization` (and then I would not need to add the parameter) but as this line was already implemented in this form, I don't know if it would be ok.

PR #641, opened 2021-05-23
A detailed CSV of the organization

### LINKED ISSUE #596: Downloadable organization data
Some organizations want to analyze their data, in past we proposed to extend reports #259

Also in #594 we add some fields to these reports

These reports are commonly used only for printing in order to facilitate participation in TimeBanks for people that have difficulties using the TimeOverflow application (old people, for example)

So maybe in the future we could keep this reports simple (printable version) and add another option for admins like "download organization Data" .zip with all CSV of data of your Organization: Users/Posts/Categories/Tags/Transactions... with all fields in use avoiding those that are only used internally in the app and do not provide data for analysis (such as password, tokens, ...) 

**PLAN**
- [x] Add the missing reports, like Transfers (👉🏼  #599)
- [ ] Add more and missing data in CSVs
- [x] Add a new option `Download all` under the hood we only need to call each CSV Report class and create a ZIP with all generated files (👉🏼  #620)

franpb14:
As agreed with @sseerrggii, the administrator can download a detailed CSV and will be able to obtain all the information shown in [this table](https://docs.google.com/document/d/1s1kVx7UVZ02EqiyBZA827cGYHFUKIddPCUWn9Tj166U/edit?usp=sharing). The "locale" and "balance" attributes of members and the "id" attribute of posts have also been deleted from the "friendly" reports.

Here you can see a demo.

https://user-images.githubusercontent.com/70280187/119259899-a45bbf80-bbd0-11eb-83ea-b22ce2dcc456.mp4

closes #596.

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -24,13 +24,17 @@ class ReportsController < ApplicationController
     report_responder('Transfer', current_organization, @transfers)
   end
 
+  def download_csv_detailed
+    download_report("Csv::Detailed", current_organization)
+  end
+
   def download_all
     filename = "#{current_organization.name.parameterize}_#{Date.today}.zip"
     temp_file = Tempfile.new(filename)
 
     begin
       Zip::File.open(temp_file.path, Zip::File::CREATE) do |zipfile|
-        %w(Member Transfer Inquiry Offer).each do |report_class|
+        %w(Member Transfer Inquiry Offer Detailed).each do |report_class|
           add_csv_to_zip(report_class, zipfile)
         end
       end
@@ -85,6 +89,8 @@ class ReportsController < ApplicationController
 
     report = if report_class.in? %w(Inquiry Offer)
       get_report("Csv::Post", current_organization, collection, report_class.constantize)
+    elsif report_class == "Detailed"
+      get_report("Csv::#{report_class}", current_organization)
     else
       get_report("Csv::#{report_class}", current_organization, collection)
     end
--- a/app/decorators/report/detailed_decorator.rb
+++ b/app/decorators/report/detailed_decorator.rb
@@ -1,0 +1,135 @@
+class Report::DetailedDecorator
+  include TransfersHelper
+
+  def initialize(org)
+    @org = org
+  end
+
+  def name(extension)
+    "#{@org.name.parameterize}_"\
+    "#{Date.today}."\
+    "#{I18n.t('global.detailed')}_"\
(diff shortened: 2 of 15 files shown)


markets:
@sseerrggii @franpb14 sorry 🙏🏼  but IMHO this is too much 😛  I'd try to go with a more simple approach, like adding the missing (and relevant) attributes (pretty sure we are so close to that scenario) to the current reports and leave only the "Download All" to download all reports. 

Another point, the `Account` model is very coupled to a user/member (or an organization). I'd say this is an implementation detail, but probably isn't necessary to expose it to the final user, so having for example the `balance` listed in the Members report makes sense from a user POV. Actually, the `Account` concept doesn't appear in the current app.

To summarize, my proposal is: tune (add/remove/update) current reports and 🚀 . @sseerrggii what attributes are you missing from current reports?

What do you think guys?

markets:
Yes @franpb14 and thank you so much for working on this!

@sseerrggii do you think we can simplify this task as per my comments above https://github.com/coopdevs/timeoverflow/pull/641#issuecomment-861783651? 🙏🏼 

I really think current (this branch I mean) implementation overlaps a bit with current functionality and exposes implementation details. Maybe am I missing something?

sseerrggii:
Iep! The problem is the purpose of reports and detailed CSV are different. 

Admins use reports for print members/offers and give the printed version to users that have some troubles to use the platform, like older people. So we want to keep this reports really simple and comprehensible for this purpose. 

Insted admins will use detailed CSV to extract information about their organization that now they can't get from statistics section, for example: 

* Last month transfers grouped by gender
* New members of last year grouped by age
* Count of groupal offers of last month
* etc

For these reasons we thought on this solution, like "export your organization data" and do you own analysis outside the app

Anyway we could live without this feature :wink: there are a few admins that asked for that... So it's you decision @markets 

From my point of view it's a way to keep the app simple and offer the option to export all data of your organization for organizations who want to do something with their data.

markets:
ah! ok @sseerrggii now I understand better the problem we want to solve here...

To be honest, I still think we can "solve" both features using current solution. I mean, I'm almost sure we already print all relevant attributes. We can add the missing ones and the reports still are (IMHO) "understandable" for people with a less technology background.

sseerrggii:
OK! What do you think if we close this PR for the moment and think in a better solution for the future?

markets:
Yes! Sorry sorry 🙏🏼  there is too much tech debt in that branch. Let's think in a simpler way!

PR #645, opened 2021-06-15
Add avatar file size validation (max 5 MB)

### LINKED ISSUE #644: Limit profile image size to 5MB
I have noticed that when you exceed the maximum file size when changing profile image, nginx gives an error "413 Request Entity Too Large"

It would be better don't allow uploading files greater than 5MB to avoid this error appearing 

Screenshot 2021-06-15 at 17-04-49 413 Request Entity Too Large

markets:
Closes #644 

Finally, instead of extra js and new gems, I implemented the size validation using 10 lines of Ruby ✨  (see `validate_avatar(file)`), following the current approach. It also includes i18n warning: 

Captura de pantalla 2021-06-16 a las 1 42 15

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -66,14 +66,12 @@ class UsersController < ApplicationController
   end
 
   def update_avatar
-    avatar = params[:avatar]
-    errors = validate_avatar(avatar)
+    operation = AvatarGenerator.new(current_user, params)
 
-    if errors.blank?
-      current_user.avatar.purge if current_user.avatar.attached?
-      crop_image_and_save(current_user, avatar)
+    if operation.errors.empty?
+      operation.call
     else
-      flash[:error] = errors.join("<br>")
+      flash[:error] = operation.errors.join("<br>")
     end
 
     redirect_to current_user
@@ -129,39 +127,5 @@ class UsersController < ApplicationController
                                  uid: id,
                                  name: @user.username)
     end
-  end
-
-  def crop_image_and_save(user, avatar)
-    orig_width = params[:original_width].to_i
-    width      = params[:height_width].to_i
-    left       = params[:width_offset].to_i
-    top        = params[:height_offset].to_i
-
-    image_processed = ImageProcessing::MiniMagick.
-                      source(avatar.tempfile).
-                      resize_to_fit(orig_width, nil).
-                      crop("#{width}x#{width}+#{left}+#{top}!").
-                      convert("png").
-                      call
-
-    user.avatar.attach(
-      io: image_processed,
-      filename: user.username,
-      content_type: avatar.content_type
-    )
-  end
-
-  def validate_avatar(file)
-    errors = []
-
-    if User::AVATAR_CONTENT_TYPES.exclude?(file.content_type)
-      errors << t("users.show.invalid_format")
-    end
-
-    if file.size.to_f > User::AVATAR_MAX_SIZE.megabytes
-      errors << t("users.avatar.max_size_warning", size: User::AVATAR_MAX_SIZE)
-    end
-
-    errors
   end
 end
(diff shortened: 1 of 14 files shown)


app/controllers/users_controller.rb
```diff
@@ -151,7 +151,17 @@ def crop_image_and_save(user, avatar)
     )
   end
 
-  def content_type_permitted(avatar_content_type)
-    %w[image/jpeg image/pjpeg image/png image/x-png].include? avatar_content_type
+  def validate_avatar(file)
```
sauloperez, line 154: have you considered moving this to a Validator class? this we wouldn't clutter the controller with a new responsibility and it could be tested in isolation.

markets: Hi @sauloperez, yes I considered moving it to validators (via https://github.com/igorkasyanchuk/active_storage_validations), but it needs more "rewrites", as the endpoint that crops and prepares the image is not a regular `user.update(params)`. In fact, we don't save the user object at all. 

Probably a good compromise is to move all this avatar operations (2 methods in this controller) to a service class, ala: `AvatarGenerator`.

sauloperez: definitely, the 1st step is to move this out of the controller. Also, we could also come up with an `AvatarValidator` of our own implemented as a simple PORO.

markets: > the 1st step is to move this out of the controller

will do it in this branch!

@sauloperez ℹ️  for the sake of reference, there is an amazing gem to unify Services/Commands interfaces in Ruby: https://github.com/markets/serviz ✨ ✨ 😺

sauloperez: which incidentally happens to be written by you :joy: it looks nice though :ok_hand:

markets: 😂  you caught me! But I can promise is the most minimalistic interface you'll need (and only 30LOC) to homogenize all your Services:

- `#call` -> invoke
- `#success?`/ `#failure?` -> ended well or bad?
- `#result` -> get the result
- `#errors` -> get the errors

app/services/avatar_generator.rb
```diff
@@ -0,0 +1,51 @@
...
+    width      = @params[:height_width].to_i
+    left       = @params[:width_offset].to_i
+    top        = @params[:height_offset].to_i
+
+    ImageProcessing::MiniMagick.
+      source(@avatar.tempfile).
+      resize_to_fit(orig_width, nil).
+      crop("#{width}x#{width}+#{left}+#{top}!").
+      convert("png").
+      call
```
sauloperez, line 39: won't this be an expensive operation risking a time out? it should perhaps be moved to a background job in a separate PR.

markets:
@sseerrggii this is ready for testing ✔️

sseerrggii:
I deployed to staging and restart the app but I keep seeing the nginx page "413 Request Entity Too Large"

markets:
Pretty sure this is related to nginx? did you provision staging server with latest changes https://github.com/coopdevs/timeoverflow-provisioning/pull/194?

sseerrggii:
Yes, I can upload files until 5M but when I upload >5MB i get de nginx error insted of the rails alert

markets:
Ah! Yes that makes sense, since Nginx works on top of Rails, it crashes first and the request doesn't hit Rails. We'll need to increase the Nginx limit I'd say, so we can capture and handle errors from the app.

sseerrggii:
Ah! Ok, so we should limit like nginx_max_body_size: 6M 

ok @markets ?

markets:
We should add a higher value there to be able to catch those errors in Rails itself, maybe something like 15MB?

sseerrggii:
Ara si!! 

Tested :heavy_check_mark: 

Thanks again @konykon and @markets

PR #648, opened 2021-06-18
TravisCI -> GH Actions

### LINKED ISSUE #646: Migrate from TravisCI to GitHub Actions
TravisCI worked really well for years, but after all changes they made in latest times, also for OSS projects... time to say 👋🏼 

GH Actions are really cool and flexible ✨ 

Some interesting articles about Travis:
 - https://daniel.haxx.se/blog/2021/06/14/bye-bye-travis-ci/
 - https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works
 
 Things to be considered:
 - We need Postgresql 9.4
 - We'd like to keep CodeClimate Test Coverage reporting
 
Help much appreciated! 🤝

markets:
Closes #646

- [x] Make tests pass 
- [x] Migrate code coverage report
- [x] Settings clean up

**NOTES** 
- CI postgres needs to define `host` and `password` DB settings.
- `collation` and `ctype` doesn't work on CI, I kept them for current staging/pro infra.
- The original `hstore` column (for categories translations) should be migrated to `jsonb` (as we are relying on https://github.com/cfabianski/json_translate). In the future (after this PR reaches production) we can remove this extension. This fixes the error:

```
ActiveRecord::StatementInvalid: PG::UndefinedFunction: ERROR:  operator does not exist: hstore @> jsonb
LINE 1: ... one FROM "categories" WHERE ("name_translations" @> '{"es":...
```


markets:
@sauloperez Did you see any similar error in the past? I basically copy-pasted from OFN, but apparently Rails can't connect to `postgresql`... I'm exposing the service at that port 🤔  

Captura de pantalla 2021-06-18 a las 20 52 51

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-18.04
+    services:
+      postgres:
+        image: postgres:9.4
+        ports: ["5432:5432"]
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+        env:
+          POSTGRES_USER: postgres
```
sauloperez, line 14: I would rather replace it with `DATABASE_USER` so we don't need to read from an extra ENV var

sauloperez:
We can try with some more `--health-retries` and see if that makes any difference :shrug:

.github/workflows/ci.yml
markets: ah ok! was thinking the `postgres` container will take that naming, I gonna change it and try with more `retries` 🤞🏼

markets:
@sauloperez changing those 2 things (ENV var naming and the `--health-retries`), we got a different error 🤷 

> Error: Failed to initialize, postgres service is unhealthy.

sauloperez:
Oops, no. You are right, those ENV vars are defined by the PostgreSQL image. Nothing to do with the `config/database.yml` ones, and at OFN we just happen to match the default values defined in `config/database.yml`.

markets:
**UPDATE** I returned back to check an extra ENV var (I'd like to avoid infra changes/coordination) and tried a couple of tweaks more (more retries and newer PostgreSQL versions), but still no luck :( 

```
PG::ConnectionBad: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

markets:
**UPDATE** After some random back and forth, I made some progress. Now the PostgreSQL service seems to start well, but it still fails due to:

```
rails aborted!
ActiveRecord::StatementInvalid: PG::UndefinedFunction: ERROR:  operator does not exist: hstore @> jsonb
LINE 1: ... one FROM "categories" WHERE ("name_translations" @> '{"es":...
```

@sauloperez do you have any idea about this one?

sauloperez:
yes, that is using the `hstore` postgreSQL extension but it doesn't seem to be installed in CI. See: https://github.com/coopdevs/timeoverflow-provisioning/blob/c8f0d859d27b5d7f41e7f370dd6d39e36f650e95/roles/database/tasks/main.yml#L56-L61.

markets:
**UPDATE** I tried forcing the hstore installation (https://github.com/coopdevs/timeoverflow/commit/7d11937025b199b0f2a3429db430281291d942b5), but no luck :(
Anyway, I think this shouldn't be unnecessary 🤔  since we are loading the `structure.sql` file, which includes this line:

https://github.com/coopdevs/timeoverflow/blob/85dd90f689ce8831dff645d14f03a60a67181e4f/db/structure.sql#L28

@sauloperez Do you think we need something else to get the `hstore` extension working?

markets:
Yay! 🎉 is green now!

markets:
@franpb14 What do you think? Could you please review it and test it locally 🙏🏼? The most "critical" thing is the migration.

config/database.yml
```diff
@@ -1,20 +1,6 @@
...
-# See https://github.com/coopdevs/timeoverflow/wiki/Keeping-your-local-files
-# for more information
-#
-
 defaults: &defaults
   adapter: postgresql
-  username: <%= ENV['DATABASE_USER'] %>  # default is null
-  collation: 'es_ES.UTF-8'
-  ctype: 'es_ES.UTF-8'
+  username: <%= ENV['DATABASE_USER'] || ENV["POSTGRES_USER"] %>
```
franpb14, line 3: out of curiosity, do we need both ENVs? can't we call the env of gh actions the same?

markets: Unfortunately in GH Actions containers, all PostgreSQL ENV vars should be prefixed by `POSTGRES_*`: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers (this is not necessary for our app, but for the PostgreSQL image to properly start up).

franpb14: oh I see, thanks!

franpb14 (approved):
Everything is working fine after the migrate. Tests & server working 🎉 🥳

PR #661, opened 2022-10-09
[Feat] Signup users

franpb14:
WIP This feature will make it possible to signup, request to join BDTs and manage these requests for admins.


app/models/petition.rb
```diff
@@ -0,0 +1,10 @@
+class Petition < ApplicationRecord
+  belongs_to :user
+  belongs_to :organization
+
+  STATUS = {
```
markets, line 5: Maybe we can use `enum`s here https://api.rubyonrails.org/v6.1/classes/ActiveRecord/Enum.html?

app/views/organizations/index.html.erb
```diff
@@ -37,6 +38,15 @@
             <td><%= org.neighborhood %></td>
             <td><%= link_to(org.web, org.web) if org.web.present? %></td>
             <td><%= org.members.count %></td>
+            <% if current_user %>
```
markets, line 41: as per this comment https://github.com/coopdevs/timeoverflow/pull/661/files#r997411162, I'd move the condition inside the `<td>`

markets (approved):
😎 cool @franpb14! I'm going to merge it to base branch and we can keep iterating in new PRs. Thanks!!

PR #678, opened 2023-03-24
[NZ changes] reset password fixes, translatable terms, Inquiry -> Request, max hours in transfers removed

franpb14:
### Changes

- max hours (20h) in transfers (personal and group) removed
- add `devise_i18n` (devise mailers translated)
- make Terms (and all Documents) DB translatable
- "Inquiry" -> "Request"


app/views/devise/mailer/reset_password_instructions.en.html.erb
markets, line 1: @franpb14 I propose to install `devise-i18n` and remove the `app/views/devise/mailer/*` folder 🙌🏼

app/views/terms/show.html.erb
```diff
@@ -1,10 +1,10 @@
 <header class="page-header">
   <h1>
-    <%= @document.title %>
+    <%= t(@document.title) %>
```
markets, line 3: I'm still thinking on that... Relying on fixed translation keys ala `t('terms.title')` seems simpler, but then we'll "lose" this automatic check by `updated_at`:

https://github.com/coopdevs/timeoverflow/blob/1820729332b7faff535c74dc26e456cba8deae14/app/controllers/application_controller.rb#L65

The current solution will require to add new keys for any terms udpate?

Another options (we should discuss them with @sseerrggii too): 
- `json_translate` gem, as we do for categories in `translates :name`
- Go with static keys and have a CONSTANT (or similar) with the timestamps of terms udpates

config/locales/en.yml
```diff
@@ -19,7 +19,7 @@ en:
         created_at: Created
         updated_at: Updated
       inquiry:
-        is_group: Group inquiry
+        is_group: Group request
```
markets, line 22: @franpb14 could you please update this "inquiry -> request" change in localeapp: https://app.localeapp.com/projects/10031. There you can find the key (save into the ENV var `LOCALEAPP_API_KEY`) to later pull it via: `bundle exec localeapp pull`.

app/views/terms/show.html.erb
markets: @franpb14 as we discussed via chat, probably moving to DB translations, via the `json_translate` gem, seems to fit better for this use-case.

db/migrate/20230401114456_make_terms_translatable.rb
```diff
@@ -0,0 +1,15 @@
+class MakeTermsTranslatable < ActiveRecord::Migration[6.1]
+  def up
+    rename_column :documents, :title, :title_translations
```
markets, line 3: @franpb14 this migration crashes for me locally:

```
== 20230401114456 MakeTermsTranslatable: migrating ============================
-- rename_column(:documents, :title, :title_translations)
   -> 0.0069s
-- rename_column(:documents, :content, :content_translations)
   -> 0.0030s
-- change_column(:documents, :title_translations, :jsonb, {:default=>{}, :null=>false, :using=>"title_translations::jsonb"})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::InvalidTextRepresentation: ERROR:  invalid input syntax for type json
DETAIL:  Token "Terminos" is invalid.
CONTEXT:  JSON data, line 1: Terminos...
/Users/markets/workspace/timeoverflow/db/migrate/20230401114456_make_terms_translatable.rb:5:in `up'
```

I think a much robust approach is:
- create new cols
- migrate data
- remove old cols

app/helpers/active_admin_helper.rb
```diff
@@ -0,0 +1,8 @@
+module ActiveAdminHelper
+  def render_translations(attribute, joiner = " ")
```
franpb14, line 2: Why this optional parameter? Wouldn't it be better to make it " | " or not optional?

markets: Yep! we can change the default value.

markets (approved):

markets:
@sseerrggii deployed to staging 🚀

sseerrggii:
Everythink looks good! :heavy_check_mark: 

Thanks @franpb14 & @markets

PR #679, opened 2023-04-02
[UI/UX] some random small enhancements 💅

markets:
- use `badge` always for tags
- remove some hardcoded styles
- use less colors for buttons
- refactor (simplify) some HTML by using `container-fluid` in main layout

👀 [Review without whitespaces](https://github.com/coopdevs/timeoverflow/pull/679/files?diff=unified&w=1)


markets:
Tested on staging 🚀

PR #706, opened 2023-10-01
[Feature] Import posts from CSV

franpb14:
### Changes
- Working in the same way as users
- CSV without headers using these attributes user_id, type (Offer or Inquiry), title, description, category_id, created_at, start_on, end_on


markets:
Thanks @franpb14 👍🏼 

Deployed to staging @sseerrggii 👀

Issue #694, opened 2023-05-21
Session timeout message still displaying in Spanish for English locale


JohnGillanders:
For a user with English language selected, the message that appears about the session timing out still displays as Spanish (on last test, a few weeks ago):
"Tu sesión ha caducado. Por favor, regístrate nuevamente para continuar."
Please translate to English.

markets:
Fixed by #712, it will be live in next release!

JohnGillanders:
Great, thanks!Regards,John GillandersManaging DirectorTechnaturallywww.Technaturally.infoYour Life, Enriched by Technology027 486 1533On 24/11/2023, at 2:52 AM, Marc Anguera ***@***.***> wrote:﻿
Fixed by #712, it will be live in next release!

nflorentin:
I think that we could reopen this issue. 

I continue to have the problem in production. 

The problem is that I could not reproduce it on my local. I change Devise.timeout_in to few seconds. I sign in, I wait a few seconds, refresh my page, and I'm seeing the message in my selected language.

On production, I'am always having the message in spanish, and sometimes even in English, even though I've always the french language selected and I'm seeing the interface in French. That means that the problem is only present in Devise module.

nflorentin:
I understood the problem. 

Devise uses Warden under the hood, and Warden uses a middleware to check if session is valid, our controller defining the "set_locale" before_action is never reached, because Warden middleware cuts the request flow before it happens.

It seems to be a small fix but it is not so small, one solution would be to write a middleware to set the locale and insert this middleware before warden middleware in order to have I18n.locale set before warden middlware executes.

PR #712, opened 2023-11-22
fix locale was not set until terms acceptance, so the locale selected was lost

nflorentin:
`set_locale` was called after `check_for_terms_acceptance!` before action.
So the locale was never saved until terms acceptance, causing some flash messages in wrong locale (ex: timeout session).

For more clarity, I also renamed `append_before_action` into `before_action` (because it is just an alias). 

I think it solves #694


markets:
Makes sense! Thanks @nflorentin

PR #713, opened 2023-11-22
adds not null db constraint and unique index on members.member_uid

### LINKED ISSUE #311: Add not null and unique constraint in member_uid column
So far the `member_uid` has a validation at model level that ensure both: its presence, and its uniqueness by organization. This however does not prevent violations. Since this is not ensured at DB level, race conditions can lead to duplicates.

We should add a unique index plus a not null constraint on that column.

nflorentin:
Closes #311

### DIFF
--- a/db/migrate/20231120164346_add_unique_index_on_member_uid_in_members.rb
+++ b/db/migrate/20231120164346_add_unique_index_on_member_uid_in_members.rb
@@ -1,0 +1,5 @@
+class AddUniqueIndexOnMemberUidInMembers < ActiveRecord::Migration[6.1]
+  def change
+    add_index :members, [:organization_id, :member_uid], unique: true
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20231120164346_add_unique_index_on_member_uid_in_members.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexOnMemberUidInMembers < ActiveRecord::Migration[6.1]
+  def change
+    add_index :members, [:organization_id, :member_uid], unique: true
```
markets, line 3: @nflorentin why is necessary this index?

nflorentin: more about it in @sauloperez comment : https://github.com/coopdevs/timeoverflow/issues/311#issue-292798958

markets: ah right!

PR #718, opened 2023-12-07
Search improvements

### LINKED ISSUE #716: all searches should be accent insensitive
All searches should be accent insensitive.

"Sérgi" should match "Sergi" and vice-versa "Sergi" should match "Sérgi".

It concerns members, inquiries/offers and tags.

nflorentin:
fix #716 

(Offers/inquiries were already case insensitive)

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -11,7 +11,7 @@
   <div class="col-md-12">
     <%= search_form_for(@search, class: "navbar-form navbar-left", url: users_path) do |f| %>
       <div class="form-group">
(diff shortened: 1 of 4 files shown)


markets (approved):
Nice 👌🏼 

Yes, offers/inquiries search is built with the `pg_search` gem:

https://github.com/coopdevs/timeoverflow/blob/12affd6af93041c0e72a9010a0b0964b65c1f3b7/app/models/post.rb#L5-L13

For some context, this was done during the Elasticsearch ✂️ #568

PR #722, opened 2023-12-09
Upgrade to Ruby 3 💫

markets:


markets:
@microstudi I pushed some small tweaks (basically this commit: https://github.com/coopdevs/timeoverflow/commit/e90d8e215c2adc0163bc2b4494b5ae9819cab4c8), but it seems it broke the Docker build job 😅 That's because I made `:amazon` the default for pro right? Hmmm 🤔 let me know what you think, but it seems a sensible default for pro.

markets:
I just made `:amazon` the default in production: 

https://github.com/coopdevs/timeoverflow/commit/e90d8e215c2adc0163bc2b4494b5ae9819cab4c8#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eL49

In local and test, it makes sense to use local storage, but for staging, pro or any production-like env, it seems a better default I'd say. Moving local files between different servers would be a pain, but pushing them to S3 make it more flexible.

microstudi:
> I just made `:amazon` the default in production:
> 
> [e90d8e2#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eL49](https://github.com/coopdevs/timeoverflow/commit/e90d8e215c2adc0163bc2b4494b5ae9819cab4c8#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eL49)
> 
> In local and test, it makes sense to use local storage, but for staging, pro or any production-like env, it seems a better default I'd say. Moving local files between different servers would be a pain, but pushing them to S3 make it more flexible.

I see, it's ok. But because the docker compose is for testing the production docker it picks up this value. Just adding the ENV var will solve this then. Go ahead!

PR #730, opened 2024-02-02
Adds a production ready docker environment

microstudi:
This PR optimizes the application for its use in Docker images:

- [x] Unifies development/production webserver by using Puma. Makes it configurable in terms of multithreading/multiprocessing.
- [x] Removes capistrano related files
- [x] Tests the docker build process in an action 
- [x] Enables and documents (see README) ENV vars for the application
- [x] Provides a docker-compose.yml production testing file (not for real use)
- [x] Optimizes a two-step Docker image. Removes the necessity of a JS runtime after a precomilation of assets
- [x] Uses a supervisor that allows to run Sidekiq/Rails on the same Docker image (this can be configured with ENV vars)
- [x] Lints ruby files with rubocop
- [x] Fixes the use of ImageMagik in Rails 7

### Testing

Run 
```
docker compose up
```

Visit http:/localhost:3000/

 ### Env vars summary:

| ENV | Description | Default |
| --- | --- | -- |
| `RAILS_ENV` | Define the rails environment (not necessary to setup unless you have some special requirements) | `production` |
| `SECRET_KEY_BASE` | Secret key for the application, generate a new one with the command `rails secret` | |
| `DATABASE_URL` | Database URL, the format is `postgresql://user:password@host:port/database` | |
| `RAILS_SERVE_STATIC_FILES` | Tell the application to serve static files (you might want to turn this off if you are using an external web server to serve files from the `public` folder) | `true` |
| `RAILS_LOG_TO_STDOUT` | Tell the application to log to STDOUT (useful for Docker) | `true` |
| `RAILS_LOG_LEVEL` | Log level for the application (use `debug` for maximum information) | `info` |
| `RAILS_MAX_THREADS` | Maximum number of threads to use in the application (use `1` if multithreading is not desired) | `5` |
| `RAILS_MIN_THREADS` | Minimum number of threads to use in the application | `RAILS_MAX_THREADS` value |
| `WEB_CONCURRENCY` | Number of web server processes to use | `2` |
| `RUN_SIDEKIQ` | Run Sidekiq worker process in the docker instance (you might want to change this if want to run different docker instances for Sidekiq and Rails) | `true` |
| `RUN_RAILS` | Run Rails web server process in the docker instance | `true` |
| `QUEUE_ADAPTER` | Adapter to use for background jobs (currently the application is using exclusively Sidekiq, so no other options here right now) | `sidekiq` |
| `SIDEKIQ_CONCURRENCY` | Number of threads to use in Sidekiq | `5` |
| `STORAGE_PROVIDER` | Storage provider for the application (currently the application supports `local` and `amazon`) | `local` |
| `FORCE_SSL` | Force SSL connections | `false` |
| `MAIL_LINK_HOST` | Host to use in the links sent by email (use your domain without protocol `mydomain.tld`) | |
| `MAIL_LINK_PROTOCOL` | Protocol to use in the previous host defined for links sent by email | `https` |
| `SMTP_ADDRESS` | SMTP server address (ie: `smtp.mailgun.org`) | |
| `SMTP_PORT` | SMTP server port (ie: `587`) | |
| `SMTP_DOMAIN` | SMTP domain (usually the application's domain) | |
| `SMTP_USER_NAME` | SMTP username | |
| `SMTP_PASSWORD` | SMTP password | |
| `SMTP_AUTHENTICATION` | SMTP authentication method | `plain` |
| `SMTP_ENABLE_STARTTLS_AUTO` | Enable STARTTLS | `true` |
| `SMTP_OPENSSL_VERIFY_MODE` | OpenSSL verify mode | `none` |
| `AWS_ACCESS_KEY_ID` | AWS access key ID (only if `STORAGE_PROVIDER` is `amazon`) | |
| `AWS_SECRET_ACCESS_KEY` | AWS secret access key (only if `STORAGE_PROVIDER` is `amazon`) | |
| `AWS_BUCKET` | AWS bucket name (only if `STORAGE_PROVIDER` is `amazon`) | |
| `AWS_REGION` | AWS region (only if `STORAGE_PROVIDER` is `amazon`) | |
| `ADMINS` | Space separated list of emails for the superadmins (ie: `admin@timeoverflow.org` | |


markets (approved):
Thanks @microstudi! Seems fine to me ✅, at least as far as I know Docker 😅

.env.example
```diff
@@ -12,6 +12,14 @@
 DATABASE_USER=postgres
 DATABASE_NAME=timeoverflow_development
 
+#RAILS CONFIG
+RAILS_SERVE_STATIC_FILES=true
+RAILS_LOG_TO_STDOUT=true
+RAILS_LOG_LEVEL=debug
+QUEUE_ADAPTER=sidekiq
```
markets, line 19: @microstudi do we really need this to be configurable via ENV var?

Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+COPY ./app /app/app
+COPY ./bin /app/bin
+COPY ./config /app/config
+COPY ./db /app/db
+COPY ./lib /app/lib
+COPY ./public/*.* /app/public/
+COPY ./config.ru /app/config.ru
+COPY ./Rakefile /app/Rakefile
+
+# Compile assets with Webpacker or Sprockets
```
markets, line 44: We are not using Webpacker (in fact this library is deprecated), maybe we can remove references to it to avoid confussions? My idea is to move to `importmap-rails` or `jsbundling-rails` in the future.

Gemfile
```diff
@@ -26,21 +25,24 @@ gem 'sidekiq-cron', '~> 1.9.1'
 gem 'aws-sdk-s3', '~> 1.94', require: false
 gem 'image_processing', '~> 1.12'
 gem 'active_storage_validations', '~> 1.1.3'
+gem "puma", ">= 5.0.0"
```
markets, line 28: Thanks 🙏🏼

I'd like to migrate to Puma for ages 😅

Gemfile
```diff
@@ -26,21 +25,24 @@ gem 'sidekiq-cron', '~> 1.9.1'
...
 # Assets
 gem 'jquery-rails', '~> 4.4.0'
 gem 'bootstrap-sass', '~> 3.4'
 gem 'sassc-rails', '~> 2.1.2'
-gem 'uglifier', '~> 4.2.0'
 gem 'select2-rails', '~> 4.0.13'
 
+group :production do
+  # we are using an ExecJS runtime only on the precompilation phase
+  gem "uglifier", "~> 4.2.0", require: false
```
markets, line 39: I'm ok with this, but I think some parity between dev/pro is also nice 👌🏼

config/database.yml
```diff
@@ -16,12 +22,13 @@ test:
 
 staging:
   <<: *defaults
-  collation: 'es_ES.UTF-8'
-  ctype: 'es_ES.UTF-8'
+  # collation: 'es_ES.UTF-8'
+  # ctype: 'es_ES.UTF-8'
```
markets, line 26: maybe we can ✂️ this commented lines? I think that was only necessary for our current and old PostgreSQL setup

markets:
@microstudi I added some small comments 🙏🏼 and I see there is also 1 failing test: https://github.com/coopdevs/timeoverflow/actions/runs/7800180296/job/21272472836?pr=730#step:5:25. The rest seems good to me 👌🏼.

config/database.yml
microstudi: yep!

Gemfile
microstudi: I've done it this way to avoid installing node on the final docker image. In any case, Uglifier is not used in the development mode as assets are not compressed.

Gemfile
microstudi: Feline!

Dockerfile
microstudi: I've removed the comments, this is harmless anyway, just a safe guard

.env.example
microstudi: I think RAILS_LOG_LEVEL needs to stay. The rest can be hidden/hardcoded. See if you prerer it as they are now

microstudi:
@markets the failing test is due the update of the imagemagik library. I've fixed it but something happened with the upload coverage.
The rest is applied!

PR #737, opened 2024-04-10
Fix/footer links

### LINKED ISSUE #734: Fix links to community
@sseerrggii it seems that https://community.coopdevs.org is down 🤔? In that case we should think something to replace a couple of mentions: https://github.com/search?q=repo%3Acoopdevs%2Ftimeoverflow%20community.coopdevs.org&type=code, assuming we don't want the Discourse back.

sseerrggii:
Some links update 

Closes #734

### DIFF
--- a/app/views/application/_footer.html.erb
+++ b/app/views/application/_footer.html.erb
@@ -9,7 +9,7 @@
         <%= render 'application/menus/language_switcher' %>
       </div>
       <div class="col-xs-12 col-sm-5 col-md-5 col-lg-4 footer-right-col">
-        <a href="https://github.com/coopdevs/timeoverflow" target="_blank">Githuub2</a>
+        <a href="https://github.com/coopdevs/timeoverflow" target="_blank">Github</a>
       </div>
     </div>
   </div>
--- a/app/views/application/_footer.html.erb
+++ b/app/views/application/_footer.html.erb
@@ -2,15 +2,14 @@
   <div class="container">
     <div class="row">
       <div class="col-xs-12 col-sm-5 col-md-5 col-lg-4 footer-left-col">
-        <a href="http://community.coopdevs.org/" target="_blank"><%= t('layouts.application.help') %></a>
+        <a href="https://adbdt.org/contacta/" target="_blank"><%= t('layouts.application.help') %></a>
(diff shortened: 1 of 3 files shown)


sseerrggii:
@markets I'm not sure why I update also db/structure.sql

it wasn't my intention :sweat_smile:

app/views/application/_footer.html.erb
```diff
@@ -2,14 +2,13 @@
...
       <div class="col-xs-12 col-sm-5 col-md-5 col-lg-4 footer-left-col">
-        <a href="http://community.coopdevs.org/" target="_blank"><%= t('layouts.application.help') %></a>
+        <a href="https://adbdt.org/contacta/" target="_blank"><%= t('layouts.application.help') %></a>
         <%= link_to t('application.terms_conditions'), tnc_path, data: { target: '#modal', toggle: 'modal' } %>
       </div>
       <div class="col-xs-12 col-sm-2 col-md-2 col-lg-4 footer-center-col">
         <%= render 'application/menus/language_switcher' %>
       </div>
       <div class="col-xs-12 col-sm-5 col-md-5 col-lg-4 footer-right-col">
-        <a href="https://twitter.com/timeoverflow" target="_blank">Twitter</a>
```
markets, line 12: no more Twitter account?

db/structure.sql
```diff
@@ -1378,3 +1373,5 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20230401114456'),
 ('20231120164231'),
 ('20231120164346');
+
+
```
markets, line 1377: Yeah, not sure why this happens, probably due to the different PostgreSQL versions we used during the migration to the new infra... Maybe we can revert this for now? I think we should solve it separately.

app/views/application/_footer.html.erb
sseerrggii: We don't use it, so I prefer not to have this link

db/structure.sql
sseerrggii: Yes, could you do it?

markets: Could you please 🙏🏼 restore it by coping the file from develop branch?

sseerrggii: Done

PR #767, opened 2024-11-04
Add "Transfer Time" text to mobile view

### LINKED ISSUE #762: Transfer time Icon
Hi everyone,

We are writing from Girona's Timebank.  Many of our members have troubles understanding  part of TimeOverflow app or mobile website, especially when they have to transfer time, because they  don't understand the meaning of the little clock shown on the right side, in the member's profile, which is used for these transactions .  We believe it would be advisable to indicate the meaning  of this icon,  exactly as it can bee sean in the desktop version of TimoeOverlow.   

Thank you!

gmartincor:
The hidden-xs and hidden-sm classes have been removed from the corresponding HTML element to display the 'Transfer Time' text in the mobile view.

Closes #762

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -5,7 +5,7 @@
         <li>
           <%= link_to edit_user_path(@user) do %>
             <%= glyph :pencil, t('global.edit') %>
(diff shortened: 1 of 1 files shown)


markets (approved):

markets:
Thanks @gmartincor going to merge this one ✅

PR #780, opened 2025-02-22
Upgrade to Bootstrap v5

franpb14:
### Main points
- I have tracked almost all the breaking changes using this excel https://docs.google.com/spreadsheets/d/135pGhEJB-vwUCRqX88EIE9KftdF-uJAAGPMOM09SMrc/edit?gid=0#gid=0
- take special attention to [this commit](https://github.com/coopdevs/timeoverflow/pull/780/commits/eb4ce1c933406d59ce114ed46bbaa20e6d97f49f), In bootstrap 3 a fontsize of 14px was applied but after upgrading there was no font-size so the 62.5% was being applied, so the font-size was very small and I had to change not only the html but the elements using rem.
- there are some changes in javascript because it dropped jQuery
- there are some things that are not exactly like in bootstrap 3, perhaps the most obvious is the secondary button which is gray now instead of white, should we change these things so they are as before?


markets:
@franpb14 couple of minor issues I just noticed:
- there is no icons rendering in the /admin section
- the top navbar in mobile is very small, it needs more height

app/assets/fonts/glyphicons-halflings-regular.eot
markets, line 1: @franpb14 If we want Sprockets to take care of this new font files, I think we should add them to the `manifest.js`. 

Example from another app:

Captura de pantalla 2025-03-04 a las 13 40 40

app/assets/stylesheets/active_admin.scss
```diff
@@ -5,9 +5,6 @@ $table-stripe-color: #f5f5f5;
 @import "active_admin/mixins";
 @import "active_admin/base";
 @import "select2";
-@import "bootstrap-sprockets";
-@import "bootstrap/variables";
-@import "bootstrap/glyphicons";
```
markets, line 10: We should keep the icons working on the admin side, see:

Captura de pantalla 2025-03-04 a las 13 42 38

app/assets/stylesheets/application/bootstrap-classes-overrides.scss
```diff
@@ -0,0 +1,7 @@
+.btn-primary {
+  @extend .text-white;
+}
+
+.bg-info.badge {
+  color: $primary;
+}
```
markets, line 7: Can't we use sass vars to tweaks those things?

app/assets/stylesheets/application/bootstrap-overrides.scss
```diff
@@ -1,10 +1,5 @@
...
-$navbar-inverse-brand-hover-color: $palette-light-blue;
-$navbar-inverse-brand-color: $white;
-$navbar-inverse-link-color: $white;
-$navbar-inverse-link-active-color: $palette-light-blue;
-$navbar-inverse-link-active-bg: $palette-turkey;
-$navbar-inverse-link-hover-color: $palette-light-blue;
-$navbar-inverse-toggle-border-color: $white;
-$navbar-inverse-toggle-hover-bg: $palette-dark-turkey;
-$navbar-inverse-bg: $palette-turkey;
+
```
markets, line 1: ✂️ empty line

app/assets/stylesheets/application/bootstrap-classes-overrides.scss
franpb14: we have this variable `$badge-color` but it  would change the color for all badges and in `badge bg-secondary` we want it white

markets:
@franpb14 just made another round of 👀 and I think overall looks quite stable now, the only thing that it works a bit weird is the top navbar in mobile:

Captura de pantalla 2025-03-05 a las 20 03 53

- It needs more height
- At some breakpoints, like the one above ☝🏼, it's a bit broken

markets:
Looks good now 👏 I'd probably push some tweaks 💅, but later in a different branch.

I think this is ready to be merged, I'm just a bit worried about the Docker build crash. That's the same script we are using to deploy the app, so right now we cant probably deploy it 🤣

sseerrggii:
@markets should I ask PokeCode about the docker problem?

markets:
@sseerrggii Would be nice 👌🏼 if they can help us to understand why this is failing and how we can fix it.

markets:
@franpb14 @sseerrggii going to merge this now, the Docker build failure seems unrelated at all, as it's also happening in another branches. We'll need to fix that of course, but in the meantime we can continue iterating on `develop`.

@microstudi

Issue #789, opened 2025-04-03
Interaction between users from different Time Banks


gmartincor:
Currently, although there is  functionality for transfers between organizations (as seen in #638 and #639), a complete system for interaction between users from different Time Banks has not been implemented. This proposal seeks to expand on that initial functionality.

To enable this broader interaction, I think we might consider updating the application's terms and conditions. Users who previously accepted the terms might need to explicitly accept the new terms upon logging in again. These new terms would include the possibility that other users can view their profile.

Additionally, I believe it would be a good idea to add a new option in "Modify my profile" (below the "Accounts" card) for each user to decide if they want to be visible to people from other time banks.

To carry this out, I would approach it through progressive phases:

### Phase 1: Acceptance between administrators

I was thinking we could implement a friendship request system between administrators of different banks. The idea would be that only when both administrators accept would their banks be considered "friendly organizations."

We could create a new "Organizations" section in the administration panel with views for accepted, rejected, and pending requests, similar to the current requests section. I also think adding a link to "Search organizations" that would take us to the organizations view (to the right of the requests) would be useful.

In the organizations view, we could include  an "Organizations" column that would only be visible to administrators. I imagine the action buttons could be something like: "Request union", "Pending", and "Delete union".

### Phase 2: Visibility of offers between users from different banks
Once the relationship between banks is established, I think the offers search view could be modified. The idea would be to add a multi-selection dropdown to filter by time bank (placed to the left of the category filter), maintain the free text search bar, and allow all these filters to be combined.

The order of elements would be: first the free text search bar, then the time bank dropdown, followed by the category dropdown, and finally the "Search" button. I also consider it important to add security validations to verify the relationships between banks.

### Phase 3: Contact between users from different banks
I've been thinking about this, and I believe there are two alternatives we could explore:

- **Idea 1:** When selecting an offer from another bank, the offerer's data would not be initially visible and the "Transfer time" button would be locked. Instead, we could add a "Send acceptance request" button. Upon sending the request, this button would change to "Acceptance request pending" and a notice about the visibility of personal data would be displayed. The offerer would receive an email notification to accept or reject, and only upon acceptance would access to their data and the transfer button be unlocked.

[...]

PR #791, opened 2025-04-15
  Organization-based filtering for posts

gmartincor:
## Context 

https://github.com/user-attachments/assets/f00b4a63-7407-4f41-8e8a-b1094f5587aa

This functionality complements the existing category filtering and allows  filtering offers by organizations and categories at the same time .

## Changes implemented

I have added an organization filtering system in the post listing interface (both offers and inquiries) that:

1. Uses the existing application design pattern based on `has_scope` to maintain consistency and follow Rails best practices  
2. Integrates seamlessly with the existing filtering system, allowing combinations with other filters (category, text search)  
3. Correctly persists filtering parameters between requests  
4. Improves the UX by keeping the button for creating new posts visible regardless of applied filters  

At the technical level, the implementation follows the MVC architecture:

- **Model**: Addition of the `by_organization` scope in the `Post` model (base class for `Offer` and `Inquiry`)  
- **Controller**: Integration of the new scope in `PostsController`, preserving existing filtering logic  
- **View**: Modification of filtering partials to include the organization selector, with behavior consistent with the category selector  

## Benefits for users

- Users can filter content by specific organization, allowing faster access to relevant posts  
- Administrators of multiple time banks can visualize and manage posts by organization  
- Users can combine filters (e.g., category + organization) for more precise searches  

## Next steps

This implementation lays the groundwork for future related improvements:

- Saving filtering preferences per user    
- Evolving towards an advanced filtering system with more options


markets:
@sseerrggii @gmartincor @franpb14 Are we sure we want this feature? 🤔 Not sure if I get the use-case.

sseerrggii:
Hi @markets yes! @gmartincor is working on #789  
In order to facilitate exchanges between organizations we need the ability to filter ofers/demands by organization

rewritten:
I tend to consider different timebanks as separate entities (even if the are tenants in the same TO instance). Any inter-bank transaction should follow the same trust criteria whether the involved banks are hosted in the same instance or in independent instances, or even across different implementations of time banking.

Here I think we hare conflating two three different feature requests:
- making it easier to administrate more than one time bank
- perusing a specific timebank requests/offers
- possibly (and this is treated separately) exchanging between banks.

PR #796, opened 2025-05-20
 Direct Transfers Between Allied Organizations

gmartincor:
# Direct Transfers Between Allied Organizations

https://github.com/user-attachments/assets/06463d99-7bca-4cb0-9e8e-29ca4bda3649

## Description
This PR adds functionality to perform direct  transfers between torganizations that have an active alliance. 

### New Controller
- Implementation of `OrganizationTransfersController` to manage specific transfers between allied organizations.

### View Modifications
- Added "Transfer Hours" button on the organization detail view when an active alliance exists
- Created form to perform transfers between organizations
- Interface to display completed transfers with a specific label

### Transfer Helper Modification
- Implementation of the `is_bank_to_bank_transfer?` method to identify direct transfers made between organizations

### Route Configuration
- New routes for the direct transfer flow between organizations

### Tests
- Tests for the `OrganizationTransfersController` verifying permissions and correct flow
- Tests for the `TransfersHelper` with the newly implemented method

## Behavior
- Only administrators can perform direct transfers from their time bank
- Transfers are only allowed between banks that have an alliance with "accepted" status
- Transfers are identified with a special label "Bank-to-Bank Transfer"
- The balance of both organizations' accounts is updated
- No post is associated with these transfers, as they represent direct collaborations between banks

## Security
- Validations to ensure only administrators can perform direct transfers
- Verification of the existence and status of the alliance between organizations before allowing transfers
- Restriction of allowed parameters to prevent mass assignments

### DIFF
--- a/app/controllers/organization_transfers_controller.rb
+++ b/app/controllers/organization_transfers_controller.rb
@@ -52,7 +52,7 @@ class OrganizationTransfersController < ApplicationController
     alliance = current_organization.alliance_with(destination_organization)
     unless alliance && alliance.accepted?
       redirect_to organizations_path,
-                  alert: t('activerecord.errors.models.transfer.attributes.base.no_alliance_between_organizations')
+                  alert: t('transfers.cross_bank.no_alliance')
     end
   end
 end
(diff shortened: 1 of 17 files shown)


rewritten:
Since bank admins can essentially create hour balances out of thin air, why do we need this? What exact problems do we want to solve?

If we had a generic way to allow a user to spend time credit in a separate _logical_ bank (no matter if they share a database), provided a baseline trust between the banks, would we still need a "backoffice" credit transfer?

sseerrggii:
Yes I agree @rewritten 
As far I know Time Banks doesn't have this use case of time transfers between organization accounts

Issue #842, opened 2026-04-10
Locale not respected in some cases


decentral1se:
I can't seem the find this text in the codebase or in the translations 🤔 I am not sure what is going on but is this something to do with the `config.i18n.fallbacks`?

https://github.com/coopdevs/timeoverflow/blob/10608592421e986d03f8531ce38ccc700af1d6c2/config/application.rb#L17

### Password reset mail

1. I choose Dutch as my locale in the admin interface
1. I import users
1. They are successfully created with the `nl-NL` locale
1. I click on "I do not remember my password" and request a password reset mail

I receive this mail using the `es` locale:

> ¡Hola X!
> 
> Alguien ha solicitado un enlace para cambiar su contraseña, lo que se puede realizar a través del siguiente enlace.
> 
> [Cambiar mi contraseña](...)
> 
> Si usted no lo ha solicitado, por favor ignore este correo electrónico.
> 
> Su contraseña no será cambiada hasta que usted acceda al enlace y cree una nueva.

Perhaps this is related to situations like https://github.com/coopdevs/timeoverflow/issues/774.

### Timebank membership management "Create user" button

Please note, I have patched in the required translations manually. Here is the PR: https://github.com/coopdevs/timeoverflow/pull/843. I hope merging this PR (or integrating the translations in the whatever way works) will officially solve this.

`/members/manage`

<img width="275" height="139" alt="Image" src="https://github.com/user-attachments/assets/45317019-5f99-47c8-94b8-5ea57dc350a6" />

### Rails admin

However, there are many buttons and texts which are not translated in the `/admin` aream, E.g.

<img width="282" height="110" alt="Image" src="https://github.com/user-attachments/assets/e18bb438-8e08-4389-a990-15589a00ec8b" />

PR #843, opened 2026-04-10
i18n: add missing translations

decentral1se:
For https://github.com/coopdevs/timeoverflow/pull/834. Related https://github.com/coopdevs/timeoverflow/issues/842.

I have updated these on https://webtranslateit.com/projects/24871-timeoverflow.

Do I need to submit a PR to sync these changes like this?

I'm still trying to figure out the workflow :upside_down_face:

### DIFF
--- a/config/locales/nl-NL.yml
+++ b/config/locales/nl-NL.yml
@@ -2,8 +2,8 @@ nl-NL:
   active_admin:
     users:
       organization: Tijdbank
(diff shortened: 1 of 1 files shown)


markets (approved):
Yes @decentral1se that's correct. The idea is to manage always the translations in the web UI and use the cli only to pull changes.

